PR #3, opened 2018-11-30
Use logging util

msau42:


msau42:
xref: https://github.com/kubernetes-csi/external-provisioner/issues/82

Issue #8, opened 2018-12-11
Remove old registration mechanism


msau42:
We should stop supporting the old way of registering the driver by directly modifying the node object.

Have we deprecated the old plugin registration mechanism yet? cc @saad-ali

msau42:
Updating docs here: https://github.com/kubernetes-csi/node-driver-registrar/pull/13

PR #10, opened 2018-12-20
Make registration path mandatory, remove kubeconfig parameter

jarrpa:
Related #8 

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>


jarrpa:
I didn't see much documented in the README so I left it alone for now. If it should be expanded somehow, that could be a separate PR.

cmd/node-driver-registrar/node_register.go
```diff
@@ -65,87 +64,42 @@ func nodeRegister(
...
 
-		// This program is intended to run as a side-car container inside a
-		// Kubernetes DaemonSet. Kubernetes DaemonSet only have one RestartPolicy,
-		// always, meaning as soon as this container terminates, it will be started
-		// again. Therefore, this program will loop indefientley and periodically
-		// update the node annotation.
-		// The CSI driver name and node ID are assumed to be immutable, and are not
-		// refetched on subsequent loop iterations.
-		for {
-			err := getVerifyAndAddNodeId(
```
msau42: Can you remove these dead functions too?

jarrpa:
@msau42 Removed dead functions, fixed test errors.

msau42:
/lgtm
/approve

As a followup, could you cleanup vendor? I think now that these k8s dependencies are removed, we should be able to remove a lot of k8s vendored pkgs.

PR #13, opened 2018-12-27
Update documentation and argument descriptions

msau42:


README.md
```diff
@@ -1,9 +1,81 @@
 [![Build Status](https://travis-ci.org/kubernetes-csi/node-driver-registrar.svg?branch=master)](https://travis-ci.org/kubernetes-csi/node-driver-registrar)
 # Node Driver Registrar
+Latest stable image: `quay.io/k8scsi/csi-node-driver-registrar:v1.0.1`
 
-A sidecar container that
+The node-driver-registrar is a sidecar container that registers the CSI driver using the
+[kubelet plugin registration mechanism](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-registration).
```
saad-ali: How about:

> The `node-driver-registrar` is a sidecar container that registers the CSI driver with Kubelet using the [kubelet plugin registration mechanism](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-registration). 
> 
> This is necessary because Kubelet is responsible for issuing CSI `NodeGetInfo`, `NodeStageVolume`, and `NodePublishVolume` calls. In order to figure out which Unix domain socket to issue the command(s) on, Kubelet checks an internal table. The `node-driver-registrar` registers your CSI driver with Kubelet so that the unix domain socket for your driver is available to kubelet.

README.md
```diff
@@ -1,9 +1,81 @@
...
+  node-driver-registrar at
+  `/var/lib/kubelet/plugins_registry/<plugin>-reg.sock` on the Kubernetes node.
+* Socket that sidecar containers and the kubelet volume subsystem uses to
+  communicate with the CSI driver and make CSI calls. This socket is created by
+  CSI driver at `/var/lib/kubelet/plugins/<plugin>/csi.sock`. This is the socket
+  referenced by the `--csi-address` and `--kubelet-registration-path` arguments.
+
+### Required arguments
+
+* `--csi-address`: This is the path to the CSI driver's socket seen inside the
```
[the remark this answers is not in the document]
msau42: should be path to the CSI driver socket, not registration

README.md
msau42: done with some modifications to the wording

README.md
```diff
@@ -1,9 +1,94 @@
...
+
+### Required arguments
+
+* `--csi-address`: This is the path to the CSI driver socket (defined above) inside the
+  pod that the `node-driver-registrar` container will use to issue CSI
+  operations (e.g. `/csi/csi.sock`).
+* `--kubelet-registration-path`: This is the path to the CSI driver socket on
+  the host node that kubelet will use to issue CSI operations (e.g.
+  `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock).
+* `KUBE_NODE_NAME` environment variable: This is the name of the Kubernetes Node.
```
saad-ali: Maybe also add: `Note that this is NOT the path to the registration socket.`

The name is pretty confusing especially now that this is the only way to operate.

README.md
```diff
@@ -1,9 +1,94 @@
...
+
+### Required permissions
+The node-driver-registrar does not interact with the Kubernetes API, so no RBAC
+rules are needed.
+
+It does, however, need to be able to mount hostPath volumes and have the file
+permissions to:
+* read from `/var/lib/kubelet/plugins/<drivername.example.com>/`
+* write to `/var/lib/kubelet/plugins_registry/`
```
saad-ali: Based on our discussion the only thing that might be needed by `node-driver-registrar` is:

> * Read and write to the registration socket (typically in the `/var/lib/kubelet/plugins_registry/` directory)
>   * Used by the `node-driver-registrar` to register the driver with Kubelet.
> * Read access to the driver socket (typically in the `/var/lib/kubelet/plugins/<drivername.example.com>/` directory)
>   * Used by the `node-driver-registrar` to fetch the driver name from the driver container (via a `GetPluginInfo(...)` CSI call).

PR #15, opened 2018-12-27
Remove retrieval of node name

msau42:
It's not needed anymore since we don't modify the Kubernetes node object


saad-ali:
Good cleanup

/lgtm
/approve

PR #29, opened 2019-02-08
node driver registrar should use CSI connection lib

andrewsykim:
re: https://github.com/kubernetes-csi/csi-lib-utils/issues/10 

Mostly followed https://github.com/kubernetes-csi/external-attacher/pull/123 with the exception of adding the reconnect callback on connection loss. Left it out because it doesn't handle address starting with `unix://` so the unit tests would fail:
```
--- FAIL: TestGetNodeID (0.00s)
    connection_test.go:90: OnConnectionLoss callback only supported for unix:// addresses
```
Left that option out for now. If it's important I can refactor the tests to use a different address? 

cc @msau42 @jsafrane

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -38,7 +38,7 @@
 
 [[constraint]]
   name = "github.com/kubernetes-csi/csi-lib-utils"
-  version = "0.1.0"
+  version = "0.3.0"
 
 [prune]
   non-go = true
--- a/cmd/csi-node-driver-registrar/main.go
+++ b/cmd/csi-node-driver-registrar/main.go
@@ -43,7 +43,7 @@ const (
 
 // Command line flags
 var (
-	connectionTimeout       = flag.Duration("connection-timeout", 1*time.Minute, "Timeout for waiting for CSI driver socket.")
+	connectionTimeout       = flag.Duration("connection-timeout", 0, "The --connection-timeout flag is deprecated")
 	csiAddress              = flag.String("csi-address", "/run/csi/socket", "Path of the CSI driver socket that the node-driver-registrar will connect to.")
 	kubeletRegistrationPath = flag.String("kubelet-registration-path", "", "Path of the CSI driver socket on the Kubernetes host machine.")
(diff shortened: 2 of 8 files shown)


Gopkg.toml
```diff
@@ -38,7 +38,7 @@
 
 [[constraint]]
   name = "github.com/kubernetes-csi/csi-lib-utils"
-  version = "0.1.0"
+  version = "0.3.0"
```
msau42: do we want to constrain the version here or just use the latest?

pkg/connection/connection.go
```diff
@@ -66,38 +63,6 @@ func NewConnection(
...
-		}
-		if conn.GetState() == connectivity.Ready {
-			klog.V(3).Infof("Connected")
-			return conn, nil
-		}
-		klog.V(4).Infof("Still trying, connection is %s", conn.GetState())
-	}
-}
-
 func (c *csiConnection) GetDriverName(ctx context.Context) (string, error) {
```
msau42: @jsafrane a lot of these rpc wrappers are also shared between sidecars. Do you think we should extract these out into csi-lib-utils too?

msau42:
> with the exception of adding the reconnect callback on connection loss

@jsafrane @pohly iiuc, the connection library will be default automatically retry connecting if the connection is lost.  So I think it's fine for now to not have any special reconnect handling for node-driver-registrar.  If we want to cache something in the future we could reconsider it.

pkg/connection/connection.go
jsafrane: I'd clean up the wrappers first, is seems to me they are copied but only GetDriverName is really used.

cmd/csi-node-driver-registrar/main.go
```diff
@@ -43,7 +43,7 @@ const (
 
 // Command line flags
 var (
-	connectionTimeout       = flag.Duration("connection-timeout", 1*time.Minute, "Timeout for waiting for CSI driver socket.")
```
mowangdk, line 46: Is there anyone knows why disable timeout here?

PR #33, opened 2019-02-26
Use GetDriverName from csi-lib-utils

andrewsykim:
/assign @msau42 @jsafrane


Gopkg.toml
```diff
@@ -19,26 +19,17 @@
...
-
-
 [[override]]
   name = "github.com/json-iterator/go"
   version = "1.1.4"
 
 [[constraint]]
   name = "github.com/kubernetes-csi/csi-lib-utils"
-  version = "0.3.0"
+  version = "0.4.0-rc1"
```
msau42: @jsafrane @pohly what should our policy be for allowing constraints on rc tags?  Will the work to have a periodic job that tries to update vendor help with automatically keeping dependencies up to date?

pohly: The automatic job (https://github.com/kubernetes-csi/csi-release-tools/issues/7) will only be able to update dependencies within the range allowed by the constraints specified in the `Gopkg.toml`. That's why I have been arguing against being overly conservative with those constraints, like locking onto exactly one revision or a too small range.

`version = "0.3.0"` was already problematic in that regard, because it gets interpreted as `>= 0.3.0, <0.4.0` and thus would have prevented an automatic update to `0.4.0` (https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#version-rules).

I'm less sure about `0.4.0-rc1`. I'm worried that it might not be interpreted as a semver by dep at all because of the `-rc1` suffix, in which case it would always pick exactly that version. But actually it's fine, https://github.com/Masterminds/semver#working-with-pre-release-versions explicitly mentions how pre-release versions like this are handled.

For `csi-lib-utils` I suggest that we use `version = ">=0.4.0-rc1"`. That allows updating to `1.0`, should we ever release that, while allowing dep to use the `rc1` (it would be ignored otherwise) and requiring it (can't use `0.3.0`).

msau42: There's already going to be breaking changes between 0.3 and 0.4. But I didn't want to bump a whole major version to put us at a 1.0 yet. Maybe for <1.0 we should not expect backwards compatibility?

pohly: I agree, there needs to be no guarantee of backwards compatibility. I would allow dep to update to any future version nevertheless, otherwise we'll get stuck on older versions because I don't expect that maintainers will follow csi-lib-utils development closely enough to notice when they need to bump a fixed version manually.

andrewsykim: Sounds good to me. So for the consistent sidecars effort, I will always update csi-lib-utils to latest, fix any breaking changes as we go and worry about backwards compatibility when we decide on a v1.0.

msau42: How about using the `^` operator? That allows it to be updated within the major version?

andrewsykim: I _think_ `^` only rounds to the [left most non-zero version](https://golang.github.io/dep/docs/Gopkg.toml.html#version). 

So `^0.4.0-rc1` would be `0.4.0-rc1` < `0.5.0`, maybe that's fine?

Or maybe we can do `version = "<1.0.0"`?

msau42: Oh I guess then @pohly 's suggestion of >= may work better for now.  We can consider restricting it to major once we have a 1.0

andrewsykim: done

PR #38, opened 2019-04-09
Add Changelog for 1.1.0

jsafrane:
@msau42, PTAL

### DIFF
--- a/CHANGELOG-1.1.md
+++ b/CHANGELOG-1.1.md
@@ -1,0 +1,18 @@
+# Changelog since v1.0.2
+
+## Deprecations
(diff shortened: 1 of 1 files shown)


msau42:
This is the commit that I think may be a breaking change: https://github.com/kubernetes-csi/node-driver-registrar/commit/6d5882517bf71826d1f031c0ba2b45aed9ee6794

jsafrane:
> This is the commit that I think may be a breaking change: 6d58825

`--kubeconfig` was already removed in v1.0.2 by https://github.com/kubernetes-csi/node-driver-registrar/pull/14

PR #47, opened 2019-08-14
Build Node Driver Registrar for Windows nodes

ddebroy:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

This PR adds support to build the Node Driver Registrar as a Windows binary: `csi-node-driver-registrar.exe`. The goal is to be able to build and publish container images for Windows using the Dockerfile.Windows. Daemonsets for CSI node plugins targeting Windows nodes can specify the Windows container packaging the Node Driver Registrar as a side-car just like in Linux.

Fixes #

The changes and build infra necessary to be able to build the Windows containers packaging `csi-node-driver-registrar.exe` is not yet present.

```release-note
Support Node Driver Registrar on Windows nodes
```

### DIFF
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,0 +1,9 @@
+FROM mcr.microsoft.com/windows/servercore:1809 as core
+
+FROM mcr.microsoft.com/windows/nanoserver:1809
+LABEL description="CSI Node driver registrar"
+
+COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
+COPY --from=core /Windows/System32/netapi32.dll /Windows/System32/netapi32.dll
+USER ContainerAdministrator
+ENTRYPOINT ["/csi-node-driver-registrar.exe"]
--- a/cmd/csi-node-driver-registrar/util_linux.go
+++ b/cmd/csi-node-driver-registrar/util_linux.go
@@ -1,0 +1,27 @@
+// +build linux
+
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func umask(mask int) (int, error) {
+	return unix.Umask(mask), nil
+}
--- a/cmd/csi-node-driver-registrar/util_windows.go
+++ b/cmd/csi-node-driver-registrar/util_windows.go
@@ -1,0 +1,27 @@
+// +build windows
+
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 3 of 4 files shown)


ddebroy:
Not sure why the changes to the build.make file are not allowed under sub-tree `release-tools`.

cmd/csi-node-driver-registrar/util_windows.go
```diff
@@ -0,0 +1,27 @@
+// +build windows
+
+/*
+Copyright 2018 The Kubernetes Authors.
```
msau42: 2019

release-tools/build.make
```diff
@@ -63,6 +63,7 @@ endif
 build-%:
 	mkdir -p bin
 	CGO_ENABLED=0 GOOS=linux go build -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$* ./cmd/$*
+	CGO_ENABLED=0 GOOS=windows go build -a -ldflags '-X main.version=$(REV) -extldflags "-static"' -o ./bin/$*.exe ./cmd/$*
```
msau42: do we also need to modify container and push rules?

release-tools/build.make
```diff
@@ -63,6 +63,7 @@ endif
 build-%:
```
msau42: Any changes to files under release-tools need to be made in https://github.com/kubernetes-csi/csi-release-tools, and then merged in the respective repos.  It's fine to make initial changes here for testing purposes, but once we determine it's good, then we need to make the PR against csi-release-tools first.

release-tools/build.make
ddebroy: Yes. That is the part that needs a bit of discussion and research to get a Windows build platform in the picture since the Windows containers can only be built on Windows hosts.

msau42:
This lgtm. Go ahead and send a PR to csi-release-tools with the windows build rule

cmd/csi-node-driver-registrar/util_linux.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package main
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func umask(mask int) (int, error) {
+	return unix.Umask(mask), nil
```
jingxu97: what is purpose of error here?

ddebroy: Will remove.

ddebroy: We need the wrapper function to able to allow platform dependent building (and skip this for Windows). The error is necessary in the Windows wrapper to clearly indicate that `umask` is not implemented on Windows. So we need the error returned for the _linux case as well.

jingxu97: With this wrapper function, do you still need to check OS in the code when using umask?

	if runtime.GOOS == "linux" {
		// Default to only user accessible socket, caller can open up later if desired
		oldmask, _ = umask(0077)
	}

Dockerfile.Windows
```diff
@@ -0,0 +1,9 @@
+FROM mcr.microsoft.com/windows/servercore:1809 as core
+
+FROM mcr.microsoft.com/windows/nanoserver:1809
+LABEL description="CSI Node driver registrar"
+
+COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
+COPY --from=core /Windows/System32/netapi32.dll /Windows/System32/netapi32.dll
+USER ContainerAdministrator
```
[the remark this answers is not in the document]
ddebroy: This is needed for the node driver registrar to be able to connect to the unix domain socket surfaced by the node CSI plugin (which also needs to be started as ContainerAdministrator).

andyzhangx:
thanks, we may also make windows build in travis CI in the future, to make this driver could always run on windows

andyzhangx:
and someone could merge this PR? Thanks.

PR #51, opened 2019-10-15
Update release tools

jsafrane:
Update release tools to current master to fix builds on non-amd64 architectures.

From other features available in new release-tools:
* Update Go mod support
* Enable hostpath expansion in e2e tests

/kind cleanup

```release-note
NONE
```


pohly:
PR #52 contains the same change. It should be possible to merge both PRs without conflicts, but history will be shorter when taking just PR #52.

jsafrane:
Ack
/close

PR #52, opened 2019-10-15
replace "dep" with "go mod"

pohly:
```release-note
Migrated to Go modules, so the source builds also outside of GOPATH.
```


jsafrane:
/lgtm
/approve
Tested with both `GO111MODULE=off make` and `GO111MODULE=on make`

PR #58, opened 2019-11-21
master: update release-tools

pohly:
Commit summary:
1eaaaa1c Delete kind cluster after tests run.
f41c1351 prow.sh: also log output of system containers

```release-note
NONE
```

### DIFF
--- a/release-tools/prow.sh
+++ b/release-tools/prow.sh
@@ -580,6 +580,19 @@ EOF
     export KUBECONFIG
 }
 
+# Deletes kind cluster inside a prow job
+delete_cluster_inside_prow_job() {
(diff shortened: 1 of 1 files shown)


msau42:
There have been more changes, do you want to include those too?

pohly:
> There have been more changes, do you want to include those too?

The script from https://github.com/kubernetes-csi/csi-release-tools/issues/7 doesn't handle updating an existing PR, so no, I would prefer to merge this as-is and later to another round of updates. Also, that way the different repos remain consistent.

pohly:
> Also, that way the different repos remain consistent.

They don't need to be, it's just a bit easier to track status when they are...

PR #59, opened 2019-12-04
fix doc about csi socket path

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

/kind documentation
cc @pohly 

Fix doc about CSI socket path.
The Kubelet is still scaning `/var/lib/kubelet/plugins` recursively, and trying to call `GetInfo` on the CSI socket, which is not implemented and thus pollute the Kubelet with harmless error logs

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,9 @@ There are two UNIX domain sockets used by the node-driver-registrar:
 * CSI driver socket:
   * Used by kubelet to interact with the CSI driver.
   * Created by the CSI driver.
-  * Exposed on a Kubernetes node via hostpath somewhere other than the Kubelet.
-    plugin registry (typically `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`).
+  * Exposed on a Kubernetes node via hostpath somewhere other than the Kubelet plugin registry. (typically `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`).
   * This is the socket referenced by the `--csi-address` and `--kubelet-registration-path` arguments.
+  * Note that before Kubernetes v1.17, if the csi socket is in the `/var/lib/kubelet/plugins/` path, kubelet may log a lot of harmless errors regarding grpc `GetInfo` call not implemented (fix in kubernetes/kubernetes#84533). The `/var/lib/kubelet/csi-plugins/` path is preferred in Kubernetes versions prior to v1.17.
 
 ### Required arguments
 


README.md
```diff
@@ -36,8 +36,7 @@ There are two UNIX domain sockets used by the node-driver-registrar:
 * CSI driver socket:
   * Used by kubelet to interact with the CSI driver.
   * Created by the CSI driver.
-  * Exposed on a Kubernetes node via hostpath somewhere other than the Kubelet.
-    plugin registry (typically `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`).
+  * Exposed on a Kubernetes node via hostpath somewhere other than the Kubelet plugin registry. (typically `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock` though a path outside of `/var/lib/kubelet/plugins/` will avoid to yield harmless error messages on the kubelet, for instance `/var/lib/kubelet/csi-plugins/<drivername.example.com>/csi.sock`).
```
pohly: The "typically `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`" is what we want to get rid of. I think it's better to not mention this at all or put it into a footnote ("For historic reasons, some CSI drivers use `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`. This has the disadvantage that kubelet will attempt to connect to a plugin via this socket, which is not supported by a CSI driver and which therefore will lead to warning messages in the kubelet log.")

Sh4d1: Hmm from what I understand, kubelet is leaning towards not scanning subdirectories in the plugins dirs, which would then eliminate this problem. Apart from that, are there any other reasons for moving the socket out of the current dir? If so any official dir?

pohly: > Hmm from what I understand, kubelet is leaning towards not scanning subdirectories in the plugins dirs, which would then eliminate this problem.

If that's the case, then why did @Madhu-1 get the following error?
```
Oct 31 13:57:11 foobar-csi-test-minion-group-xxx kubelet[1987]: E1031 13:57:11.226647    1987 plugin_watcher.go:115] error failed to get plugin info using RPC GetInfo at socket /var/lib/kubelet/plugins/com.google.csi.gcepd/csi.sock, err: rpc error: code = Unimplemented desc = unkno
wn service pluginregistration.Registration when handling create event: "/var/lib/kubelet/plugins/com.google.csi.gcepd/csi.sock": CREATE
```

> Apart from that, are there any other reasons for moving the socket out of the current dir? If so any official dir?

I'm not aware of any official recommendation. The closest to that is probably the documentation that you are currently changing...

Sh4d1: > If that's the case, then why did @Madhu-1 get the following error?
I guess because it's not implemented yet :sweat_smile: 

But yep, I'll change the wording

Sh4d1:
@pohly fixed :)

README.md
```diff
@@ -36,9 +36,9 @@ There are two UNIX domain sockets used by the node-driver-registrar:
 * CSI driver socket:
   * Used by kubelet to interact with the CSI driver.
   * Created by the CSI driver.
-  * Exposed on a Kubernetes node via hostpath somewhere other than the Kubelet.
-    plugin registry (typically `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`).
+  * Exposed on a Kubernetes node via hostpath somewhere other than the Kubelet plugin registry. (typically `/var/lib/kubelet/csi-plugins/<drivername.example.com>/csi.sock`).
   * This is the socket referenced by the `--csi-address` and `--kubelet-registration-path` arguments.
+  * Note that if the csi socket is in the `/var/lib/kubelet/plugins/` path, kubelet may log a lot of harmful errors regarding grpc `GetInfo` call not implemented (see kubernetes/kubernetes#70485)
```
msau42: This is resolved in 1.17.  We have removed CSI 0.3 support and therefore also removed watching "/var/lib/kubelet/plugins" for plugin registration sockets.  So from 1.17+, the "/var/lib/kubelet/plugins" directory won't cause log spam.

Also, I don't think "harmful" is right. It's not actually causing anything to error.

msau42: See https://github.com/kubernetes/kubernetes/pull/84533

Sh4d1: Sorry I meant harmless :sweat_smile:

Sh4d1: Okay. I'll change the phrasing :)

Sh4d1: @msau42 better now?

msau42:
/lgtm
/approve
Thanks!

PR #60, opened 2019-12-16
Ensure reg.sock exists while the registrar is running

zhucan:
Signed-off-by: zhuc <zhucan.k8s@gmail.com>

**What type of PR is this?**
> /kind bug

**What this PR does / why we need it**:
Ensure reg.sock exists while the registrar is running

**Which issue(s) this PR fixes**:
Fixes #57 

**Special notes for your reviewer**:
@jingxu97 

**Does this PR introduce a user-facing change?**:
```release-note
None
```

PR #61, opened 2019-12-18
Remove registratation socket file on shutdown

### LINKED ISSUE #21: remove reg.sock on shutdown
When driver-registrar terminates, it should ensure that it removes the `*-reg.socket` file that it created. It will get overwritten when restarting, but when experimenting with different drivers or testing with driver renaming, those files accumulate under `/var/lib/kubelet/plugins/` and then plugin registration in kubelet gets slower (several minutes!). That might be something worth fixing in kubelet (not sure?), but either way it is better to clean up.

Not much can be done when the registrar gets killed with SIGKILL or crashes, but hopefully that'll be rare.

This is something I've seen with driver-registrar 0.4.1, but probably still applies.

Madhu-1:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

when terminating, remove the *-reg.socket file created by node-driver-registrar

Signed-off-by: Madhu Rajanna <madhupr007@gmail.com>

Fixes #21

```release-note
NONE
```

### DIFF
--- a/cmd/csi-node-driver-registrar/node_register.go
+++ b/cmd/csi-node-driver-registrar/node_register.go
@@ -20,7 +20,9 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/signal"
 	"runtime"
+	"syscall"
 
 	"google.golang.org/grpc"
 
@@ -36,7 +38,7 @@ func nodeRegister(
 	// as gRPC server which replies to registration requests initiated by kubelet's
 	// pluginswatcher infrastructure. Node labeling is done by kubelet's csi code.
 	registrar := newRegistrationServer(csiDriverName, *kubeletRegistrationPath, supportedVersions)
-	socketPath := fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+	socketPath := buildSocketPath(csiDriverName)
 	if err := util.CleanupSocketFile(socketPath); err != nil {
 		klog.Errorf("%+v", err)
 		os.Exit(1)
@@ -62,11 +64,29 @@ func nodeRegister(
 	// Registers kubelet plugin watcher api.
 	registerapi.RegisterRegistrationServer(grpcServer, registrar)
 
+	go removeRegSocket(csiDriverName)
 	// Starts service
 	if err := grpcServer.Serve(lis); err != nil {
 		klog.Errorf("Registration Server stopped serving: %v", err)
 		os.Exit(1)
 	}
 	// If gRPC server is gracefully shutdown, exit
+	os.Exit(0)
+}
+
+func buildSocketPath(csiDriverName string) string {
+	return fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+}
+
+func removeRegSocket(csiDriverName string) {
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGTERM)
+	<-sigc
+	socketPath := buildSocketPath(csiDriverName)
+	err := os.Remove(socketPath)
+	if err != nil && !os.IsNotExist(err) {
+		klog.Errorf("failed to remove socket: %s with error: %+v", socketPath, err)
+		os.Exit(1)
+	}
 	os.Exit(0)
 }
--- a/README.md
+++ b/README.md
@@ -76,10 +76,6 @@ the actual driver's name.
           args:
             - "--csi-address=/csi/csi.sock"
             - "--kubelet-registration-path=/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock"
-          lifecycle:
-            preStop:
-              exec:
-                command: ["/bin/sh", "-c", "rm -rf /registration/<plugin> /registration/<drivername.example.com>-reg.sock"]
           volumeMounts:
             - name: plugin-dir
               mountPath: /csi


verult:
There's also work to make the registrar reconcile socket if it doesn't exist when it should: https://github.com/kubernetes-csi/node-driver-registrar/pull/60

Madhu-1:
@msau42 CI passed PTAL

Madhu-1:
@msau42 can you please review this PR?

msau42:
/assign @verult 
I will defer review to @verult, who is also reviewing #60, which may conflict a little with this

Madhu-1:
@verult can you please review this one.

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -20,22 +20,28 @@ import (
...
+
 func nodeRegister(
 	csiDriverName string,
 ) {
 	// When kubeletRegistrationPath is specified then driver-registrar ONLY acts
 	// as gRPC server which replies to registration requests initiated by kubelet's
 	// pluginswatcher infrastructure. Node labeling is done by kubelet's csi code.
 	registrar := newRegistrationServer(csiDriverName, *kubeletRegistrationPath, supportedVersions)
-	socketPath := fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+	socketPath := fmt.Sprintf("%s/%s-reg.sock", sockerDirPath, csiDriverName)
```
verult: nit: Could you pull this into a function like `buildSocketPath(csiDriverName)`? Right now, part of the socket path is pulled out as a const but part of it isn't

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -77,3 +84,16 @@ func nodeRegister(
...
 	os.Exit(0)
 }
+
+func removeRegSocket(csiDriverName string) {
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGTERM)
+	<-sigc
+	socketPath := fmt.Sprintf("%s/%s-reg.sock", sockerDirPath, csiDriverName)
+	err := os.Remove(socketPath)
+	if err != nil && os.IsNotExist(err) {
```
verult: Do you mean `!os.IsNotExists()`?

Madhu-1: correct. fixed now

cmd/csi-node-driver-registrar/node_register.go
Madhu-1: updated now

Madhu-1:
> Sorry for the delay and thanks for the PR! Mostly LGTM.
> 
> How did you test this change?

deleted the nodeplugin daemonset which has the node-registrar container, verified the socket present on the node or not, it was deleted

Madhu-1:
@verult PTAL

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -77,3 +80,20 @@ func nodeRegister(
...
+func buildSocketPath(csiDriverName string) string {
+	return fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+}
+
+func removeRegSocket(csiDriverName string) {
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGTERM)
+	<-sigc
+	socketPath := buildSocketPath(csiDriverName)
+	err := os.Remove(socketPath)
```
kkmsft: In the main code path we remove it only if it's a socket (not sure why). Can you please clarify why we are removing in this path without doing the same check ? I think we should have a single code path for delete which is called from the sync path and from the async signal path as well.

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -77,3 +80,20 @@ func nodeRegister(
...
 	os.Exit(0)
 }
+
+func buildSocketPath(csiDriverName string) string {
+	return fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+}
+
+func removeRegSocket(csiDriverName string) {
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGTERM)
```
kkmsft: Any ideas whether this will work on windows ? @ddebroy - any ideas ?
If not can you please move this to linux utils ?

cmd/csi-node-driver-registrar/node_register.go
Madhu-1: the check is not required here, only we are going to delete the socket if it is not present we are discarding the error, we are good here

cmd/csi-node-driver-registrar/node_register.go
Madhu-1: if someone confirms it wont work in windows i will move it linunx_utils

verult: cc @jingxu97 as well

cmd/csi-node-driver-registrar/node_register.go
verult: The original check seems wrong (if the file isn't a socket, proceed and listen to it anyway...?) It should be fixed but doesn't have to be in this PR

I agree we don't need the check here - the socket should've been created by a node-driver-registrar and if something dropped a different file in its place, it did so at its own risk

Madhu-1:
resolved merge conflict and rebased. @msau42 can you please review this one, we had a approval from @verult after rebasing bot removed it.

msau42:
/assign @ddebroy @jingxu97 
to review Windows aspect

Madhu-1:
@msau42 @ddebroy @jingxu97 can we get a review on this one?

cmd/csi-node-driver-registrar/node_register.go
ddebroy: Notifications on `syscall.SIGTERM` does work for Windows: https://golang.org/pkg/os/signal/#hdr-Windows. So LGTM from a Windows perspective. cc @jingxu97

ddebroy:
Commented about the Windows sigterm notification at https://github.com/kubernetes-csi/node-driver-registrar/pull/61/files#r463444043 . @jingxu97 if you can confirm the GCE PD Windows driver is able to clean up the scoket  file properly then this will be good 2 merge.

Issue #55, opened 2019-11-08
arm build


colek42:
Arm build of this image.  Looks like multi-arch support is supported in quay as of May 2019.
https://trello.com/c/4EcAIJrd/1303-5-quayregistry-add-support-for-manifest-list-rd

Docker also supports building arm with qemu with "experimental" enabled in the .docker/config.json

ref #48

colek42:
This is a requirement for Ceph on rpi

ref: https://github.com/ceph/ceph-csi/issues/671#issuecomment-551355568

ahuffman:
Looking for aarch64 please :)

msau42:
Closing this as dup of #48

Issue #48, opened 2019-08-27
Multi-Arch build


diederichm:
Can you please build for s390x (and maybe ppc64 and ppc64le) ?

ahuffman:
Any updates on getting a multi-arch build for this?

msau42:
Work is in progress to migrate our build infrastructure to the build pipeline supported by wg-k8s-infra. Once we migrate there, then we can look into supporting multi-arch builds.

edmondsw:
@msau42 Thanks for the update. What is the projected timeline for that migration?

msau42:
It's being actively worked on as we speak. I would expect by next month we should have something.

Also for reference, here's the suggested arches to support initially:
https://github.com/kubernetes/kubernetes/blob/master/build/pause/Makefile#L36-L54

msau42:
/assign @pohly @jingxu97 
cc @vitt-bagal

edmondsw:
@msau42 excellent! Thank you!

holdenk:
So I was also curious @ruckc it looks like the latest release does not yet have a multi-arch build. I ran `holden@hkdesktop:~/repos/scalingpythonml/make_k8s_cluster$ docker buildx imagetools inspect k8s.gcr.io/sig-storage/csi-node-driver-registrar:v1.3.0` and I got back:

```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 2466,
      "digest": "sha256:37c1d9ea538b6139df2dc2807b565afed0bf47be2323ece2b83fa2d8f088ffe1"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 655317,
         "digest": "sha256:9ff2acc3204b4093126adab3fed72de8f7bbfe332255b199c30b8b185fcf6923"
      },
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 7058616,
         "digest": "sha256:d0bfaf7febc5bdba655bb83e6806ed72c9017ecef71ee98871eb0ee1ddd5fdb3"
      }
   ]
}

```

Does anyone know of v2.0 is planned to support multi-arch builds?

pohly:
> Does anyone know of v2.0 is planned to support multi-arch builds?

[...]

PR #63, opened 2020-01-16
[WIP] Support for multi-arch

colek42:
**What type of PR is this?**
/kind feature

Support CSI on multiple arch

I am proposing that we move the build logic out of the makefile and into the multi-stage Dockerfile.  This will help us support multiple archtectures easily using `docker buildx` and ensure the build environment is reproducable.

This PR is a POC on how that may work.  Additional changes to the build process will be required.  See https://hub.docker.com/r/colek42/csi-node-driver-registrar/tags for the built images

ref: #55 #48 

```release-note
Add multi-arch image support
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,16 @@
+#export DOCKER_CLI_EXPERIMENTAL=enabled
+#docker run --rm --privileged docker/binfmt:66f9012c56a8316f9244ffd7622d7c21c1f6f28d
+#docker buildx create --use --name mybuilder
+#docker buildx build -t colek42/csi-node-driver-registrar --platform=linux/arm,linux/arm64,linux/amd64 . --push
+
+FROM golang:1.13 as builder
+WORKDIR /app
+ADD . /app/
+RUN make
+
 FROM gcr.io/distroless/static:latest
 LABEL maintainers="Kubernetes Authors"
 LABEL description="CSI Node driver registrar"
 
-COPY ./bin/csi-node-driver-registrar csi-node-driver-registrar
+COPY --from=builder /app/bin/csi-node-driver-registrar csi-node-driver-registrar
 ENTRYPOINT ["/csi-node-driver-registrar"]


colek42:
i signed it

colek42:
it looks like we can use buildx with travis.  https://medium.com/@quentin.mcgaw/cross-architecture-docker-builds-with-travis-ci-arm-s390x-etc-8f754e20aaef

colek42:
ref: https://github.com/ceph/ceph-csi/issues/671

pohly:
The proposal makes sense to me in principle. But there are some details to clarify:
- Does that also work in Prow or only in TravisCI? We wanted to move away from TravisCI, it would be a step back to add more reasons why we have to use TravisCI.
- Where and how do we choose the Go version? Right now it's specified in exactly one place (release-tools/travis.yml) and more precise with major.minor.patch version than the `FROM golang:1.13` in the Dockerfile.
- Do we keep "make build" as-is (i.e. build with Go on the host) or do we replace it with building in Docker? This is mostly relevant for developers.
- Do we need a `.dockerignore` file? Without it, `ADD . /app` may end up copying quite a bit of data from an unclear work directory when invoked by a developer. In PMEM-CSI, we blacklist everything and then whitelist just the parts that are needed for building (https://github.com/intel/pmem-csi/blob/devel/.dockerignore). This is sometimes a nuisance to maintain, though.

Dockerfile
```diff
@@ -1,6 +1,16 @@
+#export DOCKER_CLI_EXPERIMENTAL=enabled
+#docker run --rm --privileged docker/binfmt:66f9012c56a8316f9244ffd7622d7c21c1f6f28d
+#docker buildx create --use --name mybuilder
+#docker buildx build -t colek42/csi-node-driver-registrar --platform=linux/arm,linux/arm64,linux/amd64 . --push
```
Madhu-1: `Docker Buildx` is included in Docker 19.03 see https://docs.docker.com/buildx/working-with-buildx/ also we need to make sure that are we able to push this image to  quay.io as csi is using quay.io

colek42: Looks like Quay supports manifest files

ref:  https://coreos.com/quay-enterprise/releases/#1.15.2
ref:  https://docs.docker.com/registry/spec/manifest-v2-2/

> * Does that also work in Prow or only in TravisCI? We wanted to move away from TravisCI, it would be a step back to add more reasons why we have to use TravisCI.

I don't understand Prow, where can I find more info about it?  If Prow has a current verion of Docker and allows us to use buildx it should work.

> * Where and how do we choose the Go version? Right now it's specified in exactly one place (release-tools/travis.yml) and more precise with major.minor.patch version than the `FROM golang:1.13` in the Dockerfile.

If we are building in the container, I think it makes sense to specify the Go version in the 
Dockerfile.  

> * Do we keep "make build" as-is (i.e. build with Go on the host) or do we replace it with building in Docker? This is mostly relevant for developers.

I always reccomend building in a container for reproducabilty.  If developers need the same functionality we can have the Makefile call `docker build` and mount a local folder into /bin

> * Do we need a `.dockerignore` file? Without it, `ADD . /app` may end up copying quite a bit of data from an unclear work directory when invoked by a developer. In PMEM-CSI, we blacklist everything and then whitelist just the parts that are needed for building (https://github.com/intel/pmem-csi/blob/devel/.dockerignore). This is sometimes a nuisance to maintain, though.

I think a blacklist makes sense.  We are using a multi-stage build which will only copy over the artifact created in the build container.

pohly: > I don't understand Prow, where can I find more info about it? If Prow has a current verion of Docker and allows us to use buildx it should work.

buildx depends on QEMU for simple multi-platform support, right? That may create additional dependencies for the host that the job runs on. Prow is the CI for Kubernetes, running on Kubernetes itself: https://github.com/kubernetes/test-infra/blob/master/prow/README.md

> If we are building in the container, I think it makes sense to specify the Go version in the
Dockerfile.

I don't mind that. I was just pointing out that if we move it, then further work is needed to adapt the other code that currently reads the Go version from travis.yml. My concern also was that 1.13 is less precise than what we are currently using. I don't know whether it matters in practice, though.

Madhu-1: currently, quay.io doesn't support multi-arch see https://github.com/ceph/ceph-csi/pull/707#issuecomment-572088862.

Pensu: @pohly Since quay doesn't support multi-arch, could there be a possibility to move to gcr?

pohly: Yes, that has been the goal for a while now. We haven't done it yet because it wasn't clear how Prow jobs would get access to the necessary credentials for pushing to gcr. I believe that has been sorted out now.

Can you check whether `docker buildx` really works in a Prow job for multiple architectures? You can change the `release-tools/build.make` for that (for example, let `test` depend on a new `test-buildx` which runs the command), then we'll see during the next `pull-kubernetes-csi-node-driver-registrar-unit` run whether it succeeds.

BenTheElder: i'd like to get multi-arch builds off of prow and onto the GCB based image auto push infra so as to not need to register quemu on the nodes (and avoid different jobs stomping it).

I'm personally too oversubscribed for this, but the CNCF infra is generally moving that way. @justaugustus and @katharine have some work here.

claudiubelu:
> 
> 
> /cc @listx @claudiubelu @BenTheElder

You'll have to specify what kind of information you need from me. :)

But I'll assume it's Windows-related. It seems that you're only taking into account the Linux container images. The Dockerfile is different due to a few reasons (see ``Dockerfile.windows``). 

And from what I understand, you want to move the build process from the host level to the container level (using ``golang:1.13`` images), all being with a single ``docker buildx`` command - from building, publishing, manifest list. I'm not sure that ``docker buildx`` can take into account multiple ``Dockerfile`` files.

As far as binary building goes, it seems that ``golang:1.13`` has support for Windows Server 2019 (1809), so we could use that to build the necessary binaries. I think we can add a Windows node to ``docker buildx``, so that it will use that node to build the appropriate image.

But there's a concern I have at the moment. Currently, from what I've seen in the ``Dockerfile.windows`` file, we're only building images for Windows Server 2019 (1809), while we currently support also support 1903 and 1909. Now there's something you'll need to know: in order to use a 1809 container image on a 1903 or 1909 node, the node will have to have Hyper-V enabled, and the containers to be run with ``--isolation-hyperv``, which is not a good assumption to make; for example, GCE doesn't have Hyper-V enabled (AFAIK), so it's not viable. Which means that most probably we'll have to build multiple container images for Windows: for 1809, 1903, 1909, which means different image bases. But I don't know if ``docker buildx`` can support such a usecase, to build images for multiple OS versions for the same os/arch.

But even so, we'll still have to make an assumption: that the Windows build node is able to build container images for 1809, 1903, and 1909, which means that it has to have Hyper-V enabled, it's docker to have ``--isolation=hyperv`` configured, **and** it's at least a 1909 node (using Hyper-V isolation, it only ensures backwards compatibility - you can use 1809, 1903, 1909 container images on a 1909 node, but you can't use a 1909 image on a 1809 node). I guess this could be the builder's responsability to ensure this, but if we're going to have an automatic image builder for this at one point, you won't be able to do it for Windows on GCE (as I said, it doesn't have Hyper-V enabled, AFAIK), and I don't think you can have multiple Windows nodes with different OS versions build the container images for their respective versions; I don't think ``docker buildx`` is able to figure out which Windows node to use for which image based on a version.

To be fair, I'm not too familiar with ``docker buildx`` and I might be wrong about some of the points above, but it might not be answer we seek to treat all the cases above.

colek42:
>  I'm not sure that `docker buildx` can take into account multiple `Dockerfile` files.

I don't think buildkit supports Windows period.  ref: **https://github.com/moby/buildkit/issues/616** Even if it does in the near future it is likely to complicate the PR significantly.  I have customers today looking for ARM CSI support.

I think the initial hurdle is moving to prowl, and gcr.  When that is taken care of adding Multiarch support will be straight forward.

1. Move images to GCR

travisghansen:
I also would like to see this for `arm` support as well. Unfortunately I don't know anything about `prowl` or how to integrate with it but otherwise am happy to help.  Thanks!

BenTheElder:
sorry but I just do not have bandwidth for this right now.

for image pushing please sync up with k8s-infra-wg and look at the image pushing jobs https://github.com/kubernetes/test-infra/tree/master/images/builder, this tooling can be used with a wg provided GCR / GCB / ... project, on which quemu etc. should work fine. Other members of the kubernetes release group have experience with this too and use this pattern for other official images.

/uncc

justaugustus:
/remove-lifecycle stale

Instructions on configuring image building/pushing via Prow + GCB can be found here: https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing

kube-cross image building is a decent example to crib from: https://github.com/kubernetes/release/tree/master/images/build/cross

Let me know if you need additional help and feel free to assign me in PRs if you need reviews! :)

claudiubelu:
If the original author can't continue this, I could add the necessary bits, like I've done here: https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/189

justaugustus:
@claudiubelu -- I'd say give it a few days to see if there's a response before taking it over.
If you do end up taking up the work, please attribute @colek42 using the `Co-authored-by:` byline --> https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

colek42:
@claudiubelu I'd like to take a shot at this.  Right now I am aiming for a Tuesday PR.  Thank you for the ref to your work.

claudiubelu:
Great! For the most part, you can build the images in the same way I did in the PR. One key aspect is the ``manifest-tool`` usage, which allows us to also add the ``os.version`` field for an image in a manifest list, which is crucial for Windows nodes that does not have Hyper-V enabled for Hyper-V isolation. Without Hyper-V Isolation, the container OS version and the host OS version needs to match in order to start the container, and in order to properly pull the right image from the manifest list, that ``os.version`` field is crucial.

Just keep in mind to also submit the PR which creates the prow job, like this one: https://github.com/kubernetes/test-infra/pull/17335

msau42:
Hey everyone, thanks for all the attention on this. We have already been actively working on this, and have some experimental builds already going in k8s-staging-csi. Please reach out to @vitt-bagal @pohly @jingxu97 and @dims to coordinate efforts.

colek42:
@msau42 could you provide a reference to the k8s-staging-csi repo please.

msau42:
We have a WIP [cloudbuild](https://github.com/kubernetes-csi/node-driver-registrar/blob/master/cloudbuild.yaml) and the [prow job](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-csi.yaml) and [prow status](https://prow.k8s.io/?job=post-node-driver-registrar-push-images)

pohly:
Can we move over tracking of multi-arch builds to this issue: https://github.com/kubernetes-csi/csi-release-tools/issues/86

The problem is not specific to `node-driver-registrar`, although that is where we are currently experimenting.

pohly:
> Can we move over tracking of multi-arch builds to this issue

On the other hand, this issue has more context. I'm undecided...

pohly:
> On the other hand, this issue has more context. I'm undecided...

This *PR*... I was confused. We really shouldn't use a PR to discuss progress that is being made elsewhere now. I suggest we close this.

colek42:
I am quite confused about who is handling what at this point. It seems like most of the work has been done.

pohly:
> I am quite confused about who is handling what at this point. It seems like most of the work has been done.

There have been various people who experimented, most recently @vitt-bagal and @namrata-ibm. I'm now trying to get it from "POC" to "usable in practice". https://github.com/kubernetes-csi/node-driver-registrar/pull/84 hopefully will be the last PR before moving the common code and files into `csi-release-tools`.

One noteworthy difference compared to this PR: we can't do multi-stage builds with native compilation for Windows, which diminishes the value of doing it for Linux because maintaining two different ways of producing images isn't nice. Therefore https://github.com/kubernetes-csi/node-driver-registrar/pull/84 keeps building binaries on the build host outside of Docker and then just puts those into the image with a simple Dockerfile + COPY.

colek42:
@pohly That seems appropriate.  Feel free to ping me if the team decides to move in another direction

claudiubelu:
> 

> 
> One noteworthy difference compared to this PR: we can't do multi-stage builds with native compilation for Windows, which diminishes the value of doing it for Linux because maintaining two different ways of producing images isn't nice. Therefore #84 keeps building binaries on the build host outside of Docker and then just puts those into the image with a simple Dockerfile + COPY.

You don't need native building in the first place. You can do something like this: https://github.com/kubernetes-sigs/secrets-store-csi-driver/blob/master/windows.Dockerfile

As you can see in the file, the RUN part is in the Linux stage, and then the binary is just copied over inside the Windows image.

PR #64, opened 2020-02-05
Fix registration socket removal on windows

kkmsft:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fix registration socket removal on windows.

Until the bug - https://github.com/golang/go/issues/33357 is fixed, os.stat wouldn't return the
right mode(socket) on windows. Hence deleting the file, without checking whether
it's a socket, on windows.

Fixes #

```release-note
NONE
```


msau42:
/ok-to-test
/assign @verult 
who has been reviewing PRs related to registration socket cleanup

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -36,23 +36,32 @@ func nodeRegister(
...
-			os.Exit(1)
-		}
-	}
-	if err != nil && !os.IsNotExist(err) {
-		klog.Errorf("failed to stat the socket %s with error: %+v", socketPath, err)
-		os.Exit(1)
-	}
 
 	var oldmask int
 	if runtime.GOOS == "linux" {
```
ddebroy: Instead of the `if` here, how about wrapping up the logic in a function along the lines of `cleanupSocketFile()` and placing that in https://github.com/kubernetes-csi/node-driver-registrar/blob/master/cmd/csi-node-driver-registrar/util_linux.go and https://github.com/kubernetes-csi/node-driver-registrar/blob/master/cmd/csi-node-driver-registrar/util_windows.go ?

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -36,23 +36,32 @@ func nodeRegister(
...
+		fi, err := os.Stat(socketPath)
+		if err == nil && (fi.Mode()&os.ModeSocket) != 0 {
```
ddebroy: Is the `(fi.Mode()&os.ModeSocket) != 0)` useful for Linux? I understand this is exiting code - so q for the maintainers.

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -36,23 +36,32 @@ func nodeRegister(
...
+			os.Exit(1)
+		}
 		// Default to only user accessible socket, caller can open up later if desired
 		oldmask, _ = umask(0077)
+	} else {
+		// Until the bug - https://github.com/golang/go/issues/33357 is fixed, os.stat wouldn't return the
+		// right mode(socket) on windows. Hence deleting the file, without checking whether
+		// its a socket, on windows.
+		klog.Warningf("Windows environment detected. Deleting the socket path: %s", socketPath)
+		if err := os.Remove(socketPath); err != nil {
```
ddebroy: You do want to `stat` first for Windows and `os.Remove` only if the path exists, right?

kkmsft: yep, will modify.

cmd/csi-node-driver-registrar/node_register.go
kkmsft: makes sense. Will update the PR

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -36,23 +36,32 @@ func nodeRegister(
...
+				os.Exit(1)
+			}
+		}
+		if err != nil && !os.IsNotExist(err) {
+			klog.Errorf("failed to stat the socket %s with error: %+v", socketPath, err)
+			os.Exit(1)
+		}
 		// Default to only user accessible socket, caller can open up later if desired
 		oldmask, _ = umask(0077)
+	} else {
```
jingxu97: is it better to use 
if runtime.GOOS == "windows" {
...
} else {
...
}

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -36,23 +36,32 @@ func nodeRegister(
...
+		// Until the bug - https://github.com/golang/go/issues/33357 is fixed, os.stat wouldn't return the
```
jingxu97: maybe put TODO: at the beginning for easy search later?

cmd/csi-node-driver-registrar/node_register.go
kkmsft: Updated the PR. PTAL.

cmd/csi-node-driver-registrar/node_register.go
kkmsft: Updated the PR. PTAL.

cmd/csi-node-driver-registrar/node_register.go
kkmsft: Added the TODO in the util_windows.go file. PTAL.

cmd/csi-node-driver-registrar/node_register.go
kkmsft: Logic moved to per os util file. so changes in this area. PTAL.

ddebroy:
Thanks KK!
/lgtm

kkmsft:
@verult - Please take a look.

cmd/csi-node-driver-registrar/node_register.go
verult: I think Deep's intention was to not have to expose an OS check here. You could have a function where the Linux implementation calls umask() and the Windows implementation does nothing

pkg/util/util_linux.go
```diff
@@ -16,12 +16,29 @@ See the License for the specific language governing permissions and
...
 )
 
-func umask(mask int) (int, error) {
+func Umask(mask int) (int, error) {
 	return unix.Umask(mask), nil
 }
+
+func CleanupSocketFile(socketPath string) error {
+	fi, err := os.Stat(socketPath)
+	if err == nil && (fi.Mode()&os.ModeSocket) != 0 {
```
verult: FYI this might be buggy: https://github.com/kubernetes-csi/node-driver-registrar/pull/61#discussion_r377326338

We should return an error if the file isn't a socket. But yeah can be a separate PR

verult: @Madhu-1 FYI. IMO this can be reused in the deletion goroutine. We don't *need* the socket check there but it doesn't hurt. If another file type is there, the socket is gone anyway.

pkg/util/util_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"k8s.io/klog"
+)
+
+func TestSocketPath(t *testing.T) {
+	// Create in the path where the test is run
+	socketPath := "reg.sock"
+	fileExists := false
+
+	// Test1: file does not exist. So clean up should be successful.
```
verult: Could you split each test case into its own function? Easier to see which has failed

pkg/util/util_test.go
```diff
@@ -0,0 +1,110 @@
...
+	err := CleanupSocketFile(socketPath)
+	if err != nil {
+		t.Fatalf("cleanup returned error: %+v", err)
+	}
+
+	// Test2: director does not exist. So clean up should be successful.
```
verult: nit: `director` -> `directory`

pkg/util/util_linux.go
kkmsft: Thank you, @verult . Will raise a shorter PR with that change right after this one gets merged.

kkmsft: For the umask will follow up with a short PR removing the checks, since that is existing logic being modified.

cmd/csi-node-driver-registrar/node_register.go
kkmsft: For the umask will follow up with a short PR removing the checks, to separate out changes to existing code unrelated to this PR's context.

pkg/util/util_test.go
kkmsft: Fixed.

pkg/util/util_test.go
kkmsft: Changed to separate tests.

kkmsft:
@verult, @ddebroy  - Addressed review comments, please take a final look. The lgtm labels got removed since I updated the PR.

I will follow up with two small PR:
1. Fixing the logic in the linux cleanup to return error if a non-socket file is found.
2. Remove the umask GOOS check in the code paths.

@msau42 - Addressed @verult and @ddebroy comments towards lgtm. Addressed comment raised by @jingxu97.  Please take a look and consider for merging.

pkg/util/util_windows.go
```diff
@@ -0,0 +1,47 @@
+// +build windows
+
+/*
+Copyright 2019 The Kubernetes Authors.
```
jingxu97: 2020?

kkmsft: Thank you, @jingxu97 - Updated to 2020.

pkg/util/util_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"net"
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"k8s.io/klog"
+)
+
+var socketPath = "reg.sock"
```
msau42: Generally we create a unique tmp directory per testcase and create test files inside there, so that we can run multiple unit tests in parallel.

kkmsft: @msau42 : Updated the PR to have tmp test directory. PTAL.

pkg/util/util_test.go
```diff
@@ -17,51 +17,64 @@ limitations under the License.
...
 	"testing"
+	"time"
 
 	"k8s.io/klog"
 )
 
-var socketPath = "reg.sock"
+var socketFileName = "reg.sock"
+
+func getRandomTestDir() string {
```
msau42: How about using this helper? https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_plugin_test.go#L45

kkmsft: Thank you, @msau42 . Used the helper function and rebased the PR. PTAL.

kkmsft:
@msau42 - have fixed the tests, go mod vendor etc. PTAL.

msau42:
Thank you!
/lgtm
/approve

PR #66, opened 2020-03-02
feat(Dockerfile.Windows): Streamline building windows image by building the exe in a builder image stage

jchauncey:
/kind feature

Creates a builder stage image which is used to produce the exe file before building the main Windows image. This allows easier integration into CI systems for producing windows images. Also parameterizes the images so we can easily churn out multiple images for different windows versions. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,9 +1,25 @@
-FROM mcr.microsoft.com/windows/servercore:1809 as core
+# Other common ones: 1803-amd64, 1903-amd64
+# full list: curl -L https://mcr.microsoft.com/v2/windows/nanoserver/tags/list
 
-FROM mcr.microsoft.com/windows/nanoserver:1809
+ARG golangTag=1.12.6-windowsservercore-1809
+ARG nanoserverTag="mcr.microsoft.com/windows/nanoserver:1809"
+ARG servercoreTag="mcr.microsoft.com/windows/servercore:1809"
+ARG goflags_vendor=""
+ARG buildVersion=""
+
+FROM golang:${golangTag} as builder
+ENV SRC_PATH="src/github.com/kubernetes-csi/node-driver-registrar"
+COPY . ${SRC_PATH}
+RUN cd $ENV:SRC_PATH ; go build ${goflags_vendor} -a -ldflags '-X main.version=${buildVersion} -extldflags "-static"' -o ./bin/csi-node-driver-registrar.exe ./cmd/csi-node-driver-registrar ; ls ./bin ;
+
+
+FROM ${servercoreTag} as core
+
+
+FROM ${nanoserverTag}
 LABEL description="CSI Node driver registrar"
-
-COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
+# COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
 COPY --from=core /Windows/System32/netapi32.dll /Windows/System32/netapi32.dll
+COPY --from=builder gopath/src/github.com/kubernetes-csi/node-driver-registrar/bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
 USER ContainerAdministrator
 ENTRYPOINT ["/csi-node-driver-registrar.exe"]
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,7 +1,7 @@
 # Other common ones: 1803-amd64, 1903-amd64
 # full list: curl -L https://mcr.microsoft.com/v2/windows/nanoserver/tags/list
 
-ARG golangTag=1.12.6-windowsservercore-1809
+ARG golangTag=1.14.0-windowsservercore-1809
 ARG nanoserverTag="mcr.microsoft.com/windows/nanoserver:1809"
 ARG servercoreTag="mcr.microsoft.com/windows/servercore:1809"
 ARG goflags_vendor=""
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -18,7 +18,6 @@ FROM ${servercoreTag} as core
 
 FROM ${nanoserverTag}
 LABEL description="CSI Node driver registrar"
-# COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
 COPY --from=core /Windows/System32/netapi32.dll /Windows/System32/netapi32.dll
 COPY --from=builder gopath/src/github.com/kubernetes-csi/node-driver-registrar/bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
 USER ContainerAdministrator
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,7 +1,7 @@
 # Other common ones: 1803-amd64, 1903-amd64
 # full list: curl -L https://mcr.microsoft.com/v2/windows/nanoserver/tags/list
 
-ARG golangTag=1.14.0-windowsservercore-1809
+ARG golangTag=1.13.3-windowsservercore-1809
 ARG nanoserverTag="mcr.microsoft.com/windows/nanoserver:1809"
 ARG servercoreTag="mcr.microsoft.com/windows/servercore:1809"
 ARG goflags_vendor=""


jchauncey:
@kkmsft this changes the dockerfile for windows to build the exe file first and then copy it into the file image.

ritazh:
cc @PatrickLang @marosset

Dockerfile.Windows
```diff
@@ -1,9 +1,25 @@
...
-FROM mcr.microsoft.com/windows/nanoserver:1809
-LABEL description="CSI Node driver registrar"
 
-COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
+FROM ${servercoreTag} as core
+
+
+FROM ${nanoserverTag}
+LABEL description="CSI Node driver registrar"
+# COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
```
kkmsft: If we are not using this, lets just remove the line.

Dockerfile.Windows
```diff
@@ -1,9 +1,25 @@
...
+ARG golangTag=1.14.0-windowsservercore-1809
+ARG nanoserverTag="mcr.microsoft.com/windows/nanoserver:1809"
+ARG servercoreTag="mcr.microsoft.com/windows/servercore:1809"
+ARG goflags_vendor=""
+ARG buildVersion=""
+
+FROM golang:${golangTag} as builder
+ENV SRC_PATH="src/github.com/kubernetes-csi/node-driver-registrar"
+COPY . ${SRC_PATH}
+RUN cd $ENV:SRC_PATH ; go build ${goflags_vendor} -a -ldflags '-X main.version=${buildVersion} -extldflags "-static"' -o ./bin/csi-node-driver-registrar.exe ./cmd/csi-node-driver-registrar ; ls ./bin ;
```
kkmsft: Currently, it's possible to build the binaries separately and then construct the container image with it. After this change, is there a way we can still do it ?

msau42:
/ok-to-test
/assign @pohly 
We already have a Makefile rule to build windows images. It seems like this Dockerfile is bypassing those? Should we try to align them to be consistent?

jchauncey:
@msau42 I cant find any makefile rule to build the windows image. Just the binary. The problem is that the makefile setup for this project doesnt work well on a windows server so using that target is pretty much a no go. By using the dockerfile to build the windows binary and image it just reduces it down to a single command. I wanted to get this dockerfile change in since I am building the windows image we will host here at microsoft. I would like to come back and make a change in the release-tools repo to add better build support for windows. But even with that I think the Dockerfile idiom for building your binary in a builder stage and then copying it out into the main image is a really good pattern to follow.

Dockerfile.Windows
jchauncey: Yes this would eliminate that in favor of of a single `docker build` command. The 1 thing I am not crazy about is piping in the `buildVersion`. I would like to find a better way to fetch that in the dockerfile but my windows scripting knowledge isnt all that great. I figured this was a good first step and as we do more of these I will find a better way to get this information at build time and I can refactor then.

msau42: This is different from how the linux Dockerfile works. I would like to keep the behavior consistent. Can we somehow make this optional, or have another Dockerfile?

msau42:
> Michelle Au I cant find any makefile rule to build the windows image. Just the binary. 

Oops sorry I meant binary. Since the new dockerfile is copying the source, shouldn't it have access to the build rules under release-tools?

Dockerfile.Windows
jchauncey: You have no managed way to build the exe other than the makefile which wont work on windows. How do you expect your users to build this image today?

jchauncey:
No makefiles dont work on windows by default.

Dockerfile.Windows
jchauncey: You will have to build this image on windows. Doing it this way provides a singular way to do that which is managed by the repo. If you require another system to build the binary before it can build the image then you need to provide a way to build that binary that is consistent with how you build the linux binary but able to do so on said operating system.

Dockerfile.Windows
```diff
@@ -1,9 +1,24 @@
-FROM mcr.microsoft.com/windows/servercore:1809 as core
+# Other common ones: 1803-amd64, 1903-amd64
+# full list: curl -L https://mcr.microsoft.com/v2/windows/nanoserver/tags/list
+
+ARG golangTag=1.14.0-windowsservercore-1809
```
pohly: Is this picking Go 1.14? I'm a bit worried that this is not what we use for Linux (yet) and that keeping this in sync will become a maintenance issue.

But without additional scripting I don't see a way around that.

Dockerfile.Windows
pohly: > ... the makefile which wont work on windows

Why? `1.14.0-windowsservercore-1809` seems to have a POSIX shell and the corresponding tools (like `ls`). Is there no POSIX-compatible make?

The current Makefile expects also "git" for determining the version, but perhaps even that can be made available?

Dockerfile.Windows
jchauncey: Yeah I wasnt sure what was installed on the build systems you were using. I assumed it would be the latest (we are using the latest 1.13 release for our internal builds here) so I just went ahead and picked 1.14. If you can tell me what release you are using Ill match it up.

jchauncey:
The Dockerfile.Windows is already there. Im just trying to provide an easier way to get the binary into said docker image. I agree that the duplication sucks but right now Im just worried about getting an image out that we can host for users. I think release-tools needs to be modified to handle building windows images but I'm not sure what that looks like yet.

Dockerfile.Windows
pohly: The canonical definition for each component is in it's .travis.yml file, which typically is a symlink to `release-tools/travis.yml`: https://github.com/kubernetes-csi/node-driver-registrar/blob/b4bdb81f33abf9eff80c194abe67c241a5900f96/release-tools/travis.yml#L9

All other places where we need to pick a Go version are also based on that definition, for example: https://github.com/kubernetes-csi/node-driver-registrar/blob/b4bdb81f33abf9eff80c194abe67c241a5900f96/release-tools/prow.sh#L94-L100

We also warn developers when they use something else: https://github.com/kubernetes-csi/node-driver-registrar/blob/b4bdb81f33abf9eff80c194abe67c241a5900f96/release-tools/verify-go-version.sh#L33-L48

pohly: > Yeah I wasnt sure what was installed on the build systems you were using.

It's the other way around :grin: In those places that we control (CI), we install what want to use, instead of using what is installed.

Developers are free to diverge from that, but at their own risk.

pohly:
> The Dockerfile.Windows is already there.

Indeed, I missed that. I'm not familiar with the plans for Windows, so I have to defer to @ddebroy for how this was meant to be handled.

Dockerfile.Windows
jchauncey: k ive got a commit coming that updates this to 1.13.3

Dockerfile.Windows
jchauncey: alright so i should have been a bit more clear. its the fact taht the make targets call out to bash scripts that make this an issue. 

```shell
PS C:\gopath\src\github.com\kubernetes-csi\node-driver-registrar> make
./release-tools/verify-go-version.sh "go"
/usr/bin/env: 'bash': No such file or directory
make: *** [release-tools/build.make:194: check-go-version-go] Error 127
```

msau42: @pohly can we decouple the "make" rule from verifying go version?

I want to share as much common config and rules as we can so that they don't get out of sync with each other.

pohly: > bash scripts that make this an issue

Is plain POSIX acceptable? The `#!/usr/bin/env bash` stanca was copied from some other scripts. `verify-go-version.sh` should also work with just `#!/bin/sh`.

> can we decouple the "make" rule from verifying go version?

We could add an env variable for suppressing it, but in this particular case the Go version check is useful because it will trigger when `golangTag` gets out of sync. I'd prefer to make the check work.

Removing it from `make build` unconditionally would defeat the purpose of the check, because then developers also no longer get the warning.

jchauncey: really it should be os dependent and just call the right script. powershell vs bash in this case

pohly: The problem then is that we have to maintain two version of this script and potentially others. I don't know powershell, so I can't create or maintain anything written in it.

POSIX has been around for a while. If there is a way to support it also for Windows builds, then I vote for enabling that and writing cleaner POSIX-compatible scripts that then also run on Windows.

jchauncey: @PatrickLang or @marosset would probably be able to shed a bit more light on this as I am definitely not a windows expert here.

jchauncey:
im just going to close this and come back to it later. I think the work should be done in release-tools to help promote more continuity between the linux builds and windows builds

PR #77, opened 2020-04-22
Make base image and tag configrable in Dockerfile.Windows

jingxu97:
This PR makes the base image and tag configrable in Dockerfile.Windows

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note
Make image, tag and registry configurable in docker file for windows build of node driver registrar.
```

### DIFF
--- a/Dockerfile.Windows
+++ b/Dockerfile.Windows
@@ -1,9 +1,13 @@
-FROM mcr.microsoft.com/windows/servercore:1809 as core
+ARG BASE_IMAGE=servercore
+ARG BASE_IMAGE_TAG=1809
+ARG REGISTRY=mcr.microsoft.com/windows
 
+FROM ${REGISTRY}/${BASE_IMAGE}:${BASE_IMAGE_TAG} as core
 FROM mcr.microsoft.com/windows/nanoserver:1809
 LABEL description="CSI Node driver registrar"
 
 COPY ./bin/csi-node-driver-registrar.exe /csi-node-driver-registrar.exe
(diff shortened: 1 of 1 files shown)


Dockerfile.Windows
```diff
@@ -1,4 +1,7 @@
-FROM mcr.microsoft.com/windows/servercore:1809 as core
+ARG BASE_IMAGE
+ARG BASE_IMAGE_TAG
+
+FROM mcr.microsoft.com/windows/${BASE_IMAGE}:${BASE_IMAGE_TAG} as core
```
msau42, line 4: Does registry also need to be configurable?

Dockerfile.Windows
```diff
@@ -1,4 +1,7 @@
...
 
 FROM mcr.microsoft.com/windows/nanoserver:1809
```
msau42, line 6: Does this base image need to be configurable?

Dockerfile.Windows
jingxu97: for now, not really. But I added it in case and set default value

Dockerfile.Windows
jingxu97: this is kind of workaround, currently we have to use nanoserver due to socket issue. In the future, we need to find a solution because microsoft will stop supporting nanoserver.

jingxu97:
@msau42 I addressed the comments. PTAL

msau42:
/lgtm
/approve

Can you also add a release note?

PR #82, opened 2020-05-12
Tagging of multiarch images on gcr

vitt-bagal:
What type of PR is this?
/kind feature

What this PR does / why we need it:
This PR will push canary image for `master` branch and `vX.Y.Z` image depending on tag.

@pohly Could you please review ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,6 @@ all: build
 
 include release-tools/build.make
 
-# Override registry, revision and Images
-REV=$(REV_CI)
-REGISTRY_NAME=$(REGISTRY_CI)
-IMAGE_NAME=$(REGISTRY_NAME)/$*
-
 # This target builds multiarch images using Moby BuildKit builder toolkit.
 # Docker Buildx is included in Docker 19.03 and needs DOCKER_CLI_EXPERIMENTAL enabled to run corresponding commands.
 # Currently amd, s390x and Windows manifest is pushed for canary, release branch and released tags.
@@ -32,15 +27,14 @@ push-multiarch-%:
 	make BUILD_PLATFORMS="windows amd64 .exe"
 	set -ex; \
 	gcloud auth configure-docker; \
-	label_rev=v$$(echo $(REV) | cut -f3- -d 'v'); \
 	DOCKER_CLI_EXPERIMENTAL=enabled; \
 	export DOCKER_CLI_EXPERIMENTAL; \
 	docker buildx create --use --name multiarchimage-buildertest; \
 	pushMultiArch () { \
                 tag=$$1; \
-                docker buildx build --push -t $(IMAGE_NAME):amd64-linux-$$tag --platform=linux/amd64 -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$$label_rev .; \
-                docker buildx build --push -t $(IMAGE_NAME):s390x-linux-$$tag --platform=linux/s390x -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$$label_rev .; \
-                docker buildx build --push -t $(IMAGE_NAME):amd64-windows-$$tag --platform=windows -f $(shell if [ -e ./cmd/$*/Dockerfile.Windows ]; then echo ./cmd/$*/Dockerfile.Windows; else echo Dockerfile.Windows; fi) --label revision=$$label_rev .; \
+                docker buildx build --push -t $(IMAGE_NAME):amd64-linux-$$tag --platform=linux/amd64 -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$(REV) .; \
+                docker buildx build --push -t $(IMAGE_NAME):s390x-linux-$$tag --platform=linux/s390x -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$(REV) .; \
+                docker buildx build --push -t $(IMAGE_NAME):amd64-windows-$$tag --platform=windows -f $(shell if [ -e ./cmd/$*/Dockerfile.Windows ]; then echo ./cmd/$*/Dockerfile.Windows; else echo Dockerfile.Windows; fi) --label revision=$(REV) .; \
                 docker manifest create --amend $(IMAGE_NAME):$$tag $(IMAGE_NAME):amd64-linux-$$tag \
                         $(IMAGE_NAME):s390x-linux-$$tag \
                         $(IMAGE_NAME):amd64-windows-$$tag; \
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,52 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+.PHONY: push-multiarch-% push-multiarch
+
 CMDS=csi-node-driver-registrar
 all: build
 
 include release-tools/build.make
+
+# Override registry, revision and Images
+REV=$(REV_CI)
+REGISTRY_NAME=$(REGISTRY_CI)
+IMAGE_NAME=$(REGISTRY_NAME)/$*
+
+# This target builds multiarch images using Moby BuildKit builder toolkit.
+# Docker Buildx is included in Docker 19.03 and needs DOCKER_CLI_EXPERIMENTAL enabled to run corresponding commands.
+# Currently amd, s390x and Windows manifest is pushed for canary, release branch and released tags.
+# Images generated from Prow build are pushed to staging area on gcr
+push-multiarch-%:
+	make BUILD_PLATFORMS="windows amd64 .exe"
+	set -ex; \
+	gcloud auth configure-docker; \
+	label_rev=v$$(echo $(REV) | cut -f3- -d 'v'); \
+	DOCKER_CLI_EXPERIMENTAL=enabled; \
+	export DOCKER_CLI_EXPERIMENTAL; \
+	docker buildx create --use --name multiarchimage-buildertest; \
+	pushMultiArch () { \
+                tag=$$1; \
+                docker buildx build --push -t $(IMAGE_NAME):amd64-linux-$$tag --platform=linux/amd64 -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$$label_rev .; \
+                docker buildx build --push -t $(IMAGE_NAME):s390x-linux-$$tag --platform=linux/s390x -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$$label_rev .; \
+                docker buildx build --push -t $(IMAGE_NAME):amd64-windows-$$tag --platform=windows -f $(shell if [ -e ./cmd/$*/Dockerfile.Windows ]; then echo ./cmd/$*/Dockerfile.Windows; else echo Dockerfile.Windows; fi) --label revision=$$label_rev .; \
+                docker manifest create --amend $(IMAGE_NAME):$$tag $(IMAGE_NAME):amd64-linux-$$tag \
+                        $(IMAGE_NAME):s390x-linux-$$tag \
+                        $(IMAGE_NAME):amd64-windows-$$tag; \
+                docker manifest push -p $(IMAGE_NAME):$$tag; \
+	}; \
+	if [ $(PULL_BASE_REF) = "master" ]; then \
+                       : "creating or overwriting canary image"; \
+                       pushMultiArch canary; \
+	elif echo $(PULL_BASE_REF) | grep -q -e 'release-*' ; then \
+                       : "creating or overwriting canary image for release branch"; \
+                        release_canary_tag=$$(echo $(PULL_BASE_REF) | cut -f2 -d '-')-canary; \
+                        pushMultiArch $$release_canary_tag; \
+	elif docker pull $(IMAGE_NAME):$(PULL_BASE_REF) 2>&1 | tee /dev/stderr | grep -q "manifest for $(IMAGE_NAME):$(PULL_BASE_REF) not found"; then \
+                       : "creating release image"; \
+                       pushMultiArch $(PULL_BASE_REF); \
+	else \
+                       : "release image $(IMAGE_NAME):$(PULL_BASE_REF) already exists, skipping push"; \
+	fi; \
+
+push-multiarch: $(CMDS:%=push-multiarch-%)
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -10,12 +10,11 @@ steps:
   - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20200421-a2bf5f8'
     entrypoint: make
     env:
-    - REV_CI=$_GIT_TAG
     - PULL_BASE_REF=$_PULL_BASE_REF
     - HOME=/root
-    - REGISTRY_CI=gcr.io/k8s-staging-csi
     args:
-    - push-multiarch
+    # Extract tag-n-hash value from _GIT_TAG(form vYYYYMMDD-tag-n-hash) for REV value
+    - push-multiarch REV=v$(echo $_GIT_TAG | cut -f3- -d 'v') REGISTRY_NAME=gcr.io/k8s-staging-csi
 substitutions:
   # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
   # can be used as a substitution
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -12,28 +12,17 @@ steps:
     env:
     - DOCKER_CLI_EXPERIMENTAL=enabled
     - VERSION=$_GIT_TAG
-    - BASE_REF=$_PULL_BASE_REF
+    - PULL_BASE_REF=$_PULL_BASE_REF
     - HOME=/root
     args:
     - '-c'
     - |
-      export DOCKER_CLI_EXPERIMENTAL=enabled \
-      && gcloud auth configure-docker \
-      && docker buildx create --use --name multiarchimage-builder \
-      && docker buildx build --push -t gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:amd64-linux-$_GIT_TAG --platform=linux/amd64 -f Dockerfile.multiarch . \
-      && docker buildx build --push -t gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:s390x-linux-$_GIT_TAG --platform=linux/s390x -f Dockerfile.multiarch . \
-      && make BUILD_PLATFORMS="windows amd64 .exe" \
-      && docker buildx build --push -t gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:amd64-windows-$_GIT_TAG --platform=windows -f Dockerfile.Windows . \
-      && docker manifest create --amend gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:$_GIT_TAG \
-        gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:amd64-linux-$_GIT_TAG \
-        gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:s390x-linux-$_GIT_TAG \
-        gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:amd64-windows-$_GIT_TAG \
-      && docker manifest push -p gcr.io/$_STAGING_PROJECT/csi-node-driver-registrar:$_GIT_TAG
+      make push-multiarch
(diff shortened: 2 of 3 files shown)


release-tools/build.make
```diff
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: build-% build container-% container push-% push clean test
+# Added push-multiarch to build and push multiarch image
```
pohly, line 15: That comment can be removed.

release-tools/build.make
```diff
@@ -52,6 +53,7 @@ IMAGE_TAGS+=$(shell tagged="$$(git describe --tags --match='v*' --abbrev=0)"; if
 
 # Images are named after the command contained in them.
 IMAGE_NAME=$(REGISTRY_NAME)/$*
+MULTIARCH_IMAGE_NAME=gcr.io/k8s-staging-csi/$*
```
pohly, line 56: I don't think we need a separate `MULTIARCH_IMAGE_NAME`. Just use `IMAGE_NAME` and make sure that `cloudbuild.yaml` overrides it with `make push-multiarch REGISTRY_NAME=gcr.io/k8s-staging-csi`.

release-tools/build.make
```diff
@@ -101,9 +106,42 @@ push-%: container-%
 		fi; \
 	done
 
+RELEASE_ALIAS_TAG=$(PULL_BASE_REF)
+
+push-multiarch-%:
+	export DOCKER_CLI_EXPERIMENTAL=enabled
+	make BUILD_PLATFORMS="windows amd64 .exe"
```
pohly, line 113: Why this line?

release-tools/build.make
```diff
@@ -101,9 +106,42 @@ push-%: container-%
 		fi; \
 	done
 
+RELEASE_ALIAS_TAG=$(PULL_BASE_REF)
+
+push-multiarch-%:
+	export DOCKER_CLI_EXPERIMENTAL=enabled
```
pohly, line 112: This doesn't do anything, the variable is getting set and the shell in which it was set exits. You need to join the lines with `; \`:
```
push-multiarch-%:
    gcloud auth configure-docker
    set -ex; \
    DOCKER_CLI_EXPERIMENTAL=enabled; \
    export DOCKER_CLI_EXPERIMENTAL; \
    docker buildx create --use --name multiarchimage-buildertest; \
    pushMultiArch () { \
```

I intentionally avoided `export DOCKER_CLI_EXPERIMENTAL=enabled` because that only works in bash.

release-tools/build.make
```diff
@@ -101,9 +106,42 @@ push-%: container-%
...
+	make BUILD_PLATFORMS="windows amd64 .exe"
+	gcloud auth configure-docker
+	docker buildx create --use --name multiarchimage-buildertest
+	set -ex; \
+	pushMultiArch () { \
+                tag=$$1  ;\
+                docker buildx build --push -t $(MULTIARCH_IMAGE_NAME):amd64-linux-$$tag --platform=linux/amd64 -f Dockerfile.multiarch . ;\
```
pohly, line 119: We loose the ability here to build different images in the same repo with different Docker files. Can you add something similar to `build-%` where it locates the Docker file with `$(shell if [ -e ./cmd/$*/Dockerfile ]; then echo ./cmd/$*/Dockerfile; else echo Dockerfile; fi)`?

release-tools/build.make
```diff
@@ -101,9 +106,42 @@ push-%: container-%
 		fi; \
 	done
 
+RELEASE_ALIAS_TAG=$(PULL_BASE_REF)
+
+push-multiarch-%:
```
pohly, line 111: This whole target needs documentation. A short explanation of how it works and what the expected environment is would be useful.

release-tools/build.make
```diff
@@ -101,9 +106,42 @@ push-%: container-%
...
+                docker buildx build --push -t $(MULTIARCH_IMAGE_NAME):s390x-linux-$$tag --platform=linux/s390x -f Dockerfile.multiarch . ;\
+                docker buildx build --push -t $(MULTIARCH_IMAGE_NAME):amd64-windows-$$tag --platform=windows -f Dockerfile.Windows . ;\
+                docker manifest create --amend $(MULTIARCH_IMAGE_NAME):$$tag $(MULTIARCH_IMAGE_NAME):amd64-linux-$$tag \
+                        $(MULTIARCH_IMAGE_NAME):s390x-linux-$$tag \
+                        $(MULTIARCH_IMAGE_NAME):amd64-windows-$$tag ;\
+                docker manifest push -p $(MULTIARCH_IMAGE_NAME):$$tag ;\
+	}; \
+	if [ $(RELEASE_ALIAS_TAG) = "master" ]; then \
+                       : "creating or overwriting canary image"; \
+                       pushMultiArch canary ; \
```
pohly, line 129: For canary images, the `rev` label used to specify the source code revision (= result of git describe). This is useful to determine what the image contains because there is no other indicator.

Can this be done also for cloud builds? If we override `REV` in `cloudbuild.yaml`, then we can simply use `--label rev=$(REV)` here.

release-tools/build.make
```diff
@@ -101,9 +106,42 @@ push-%: container-%
 		fi; \
 	done
 
+RELEASE_ALIAS_TAG=$(PULL_BASE_REF)
```
pohly, line 109: Introducing another name for `PULL_BASE_REF` IMHO obscures what's going on. Better just use `PULL_BASE_REF` directly below.

cloudbuild.yaml
```diff
@@ -12,28 +12,17 @@ steps:
     env:
     - DOCKER_CLI_EXPERIMENTAL=enabled
```
pohly, line 13: This can be removed now, right?

cloudbuild.yaml
```diff
@@ -12,28 +12,17 @@ steps:
...
     - VERSION=$_GIT_TAG
```
pohly, line 14: Instead of exporting an unused VERSION env variable, can we use this to override the useless `REV` make variable, like this:
```
make push-multiarch 'REV=$_GIT_TAG'
```

Or do substitutions only work under `env`?

release-tools/build.make
```diff
@@ -69,6 +71,9 @@ endif
 # toolchain.
 BUILD_PLATFORMS =
 
+# To enable experimental features on the Docker daemon
+export DOCKER_CLI_EXPERIMENTAL:=enabled
```
pohly, line 75: IMHO this is better done locally in push-multiarch-%.

release-tools/build.make
vitt-bagal: removed comment

release-tools/build.make
vitt-bagal: Added this in push-multiarch target

cloudbuild.yaml
vitt-bagal: yes

release-tools/build.make
vitt-bagal: Removed `RELEASE_ALIAS_TAG`. Using `PULL_BASE_REF` directly in push-multiarch target

release-tools/build.make
vitt-bagal: Updated `docker buildx build` command using approach of locating file.

release-tools/build.make
vitt-bagal: Added short explanation about this target

release-tools/build.make
vitt-bagal: Removed `MULTIARCH_IMAGE_NAME` extra variable. Added `REGISTRY_CI` env variable in cloudbuild to override `REGISTRY_NAME` value in makefile

release-tools/build.make
vitt-bagal: make build for windows is required as Windows binary is needed to copy it in windows dockerfile

release-tools/build.make
vitt-bagal: As `rev` value from git describe command(build.make) is same as `tag-n-hash` value from _GIT_TAG( vYYYYMMDD-tag-n-ghash form). Added code to extract `tag-n-hash` value from REV(REV_CI cloudbuild env)in push-multiarch target and passing this as label value to docker build.

cloudbuild.yaml
vitt-bagal: Added changes in Makefile to override REV value with REV_CI value passed as env in cloudbuild 

> do substitutions only work under env? 

I am also not sure about how substitutions works. May be you can tag someone who can comment on this

vitt-bagal:
@pohly Thanks for review comment. Added few changes as per your suggestions.

Makefile
```diff
@@ -12,7 +12,52 @@
...
+
+# This target builds multiarch images using Moby BuildKit builder toolkit.
+# Docker Buildx is included in Docker 19.03 and needs DOCKER_CLI_EXPERIMENTAL enabled to run corresponding commands.
+# Currently amd, s390x and Windows manifest is pushed for canary, release branch and released tags.
+# Images generated from Prow build are pushed to staging area on gcr
+push-multiarch-%:
+	make BUILD_PLATFORMS="windows amd64 .exe"
+	set -ex; \
+	gcloud auth configure-docker; \
+	label_rev=v$$(echo $(REV) | cut -f3- -d 'v'); \
```
pohly, line 35: Is this specific to the cloud build? Then put it into `cloudbuild.yaml` and explain what it does.

Makefile
```diff
@@ -12,7 +12,52 @@
...
+
 CMDS=csi-node-driver-registrar
 all: build
 
 include release-tools/build.make
+
+# Override registry, revision and Images
+REV=$(REV_CI)
+REGISTRY_NAME=$(REGISTRY_CI)
+IMAGE_NAME=$(REGISTRY_NAME)/$*
```
pohly, line 25: Can you do this without `REGISTRY_CI` and `REV_CI`? The goal is to write a make target that can be copied 1:1 into `release-tools/build.make`, with as few special cases as possible.

You can use `REV` and `REGISTRY_NAME` and override them with `make REV=... REGISTRY_NAME=...` in the `cloudbuild.yaml` invocation.

Makefile
vitt-bagal: Moved rev calculation to `cloudbuild.yaml`

Makefile
vitt-bagal: Removed extra CI variables. Updated push-multiarch invocation as `make push-multiarch REV=v$(echo $_GIT_TAG | cut -f3- -d 'v') REGISTRY_NAME=gcr.io/k8s-staging-csi` in cloudbuild

Makefile
```diff
@@ -12,7 +12,46 @@
...
 all: build
 
 include release-tools/build.make
+
+# This target builds multiarch images using Moby BuildKit builder toolkit.
+# Docker Buildx is included in Docker 19.03 and needs DOCKER_CLI_EXPERIMENTAL enabled to run corresponding commands.
+# Currently amd, s390x and Windows manifest is pushed for canary, release branch and released tags.
+# Images generated from Prow build are pushed to staging area on gcr
+push-multiarch-%:
+	make BUILD_PLATFORMS="windows amd64 .exe"
```
pohly, line 27: I'm still not sure what this `make` invocation is needed for.

Ah, it's because `Dockerfile.Windows` doesn't build its binary. We probably should fix this and ideally remove the special Dockerfile for Windows entirely. But this can probably wait until later. We have to sort it out before including it in csi-release-tools.

@ddebroy: is there a chance to use `gcr.io/distroless/static:latest` as base for a Windows image?

Makefile
```diff
@@ -12,7 +12,46 @@
...
+	set -ex; \
+	gcloud auth configure-docker; \
+	DOCKER_CLI_EXPERIMENTAL=enabled; \
+	export DOCKER_CLI_EXPERIMENTAL; \
+	docker buildx create --use --name multiarchimage-buildertest; \
+	pushMultiArch () { \
+                tag=$$1; \
+                docker buildx build --push -t $(IMAGE_NAME):amd64-linux-$$tag --platform=linux/amd64 -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$(REV) .; \
```
pohly, line 35: Perhaps create a local variable `dockerfile=$(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi)` above and use it here and below?

Makefile
```diff
@@ -12,7 +12,46 @@
...
+                docker buildx build --push -t $(IMAGE_NAME):s390x-linux-$$tag --platform=linux/s390x -f $(shell if [ -e ./cmd/$*/Dockerfile.multiarch ]; then echo ./cmd/$*/Dockerfile.multiarch; else echo Dockerfile.multiarch; fi) --label revision=$(REV) .; \
+                docker buildx build --push -t $(IMAGE_NAME):amd64-windows-$$tag --platform=windows -f $(shell if [ -e ./cmd/$*/Dockerfile.Windows ]; then echo ./cmd/$*/Dockerfile.Windows; else echo Dockerfile.Windows; fi) --label revision=$(REV) .; \
```
pohly, line 37: Have you tried doing this with `Dockerfile.multiarch` instead of `Dockerfile.Windows`?

You must extract the target OS from TARGETPLATFORM instead of hard-coding `linux` and the .exe suffix may need special handling.

vitt-bagal: >Have you tried doing this with Dockerfile.multiarch instead of Dockerfile.Windows?

I haven't tried this. We can't test windows images at our end. May be @jingxu97 can try this out.

Ideally instead of writing different `docker buildx build` command for each arch build and then creating manifest for them, single command with `--push --platform=linux/amd64,linux/s390x` should work. However I feel  we can try this out later, once current changes are working as expected.

Makefile
vitt-bagal: I think we can use this once windows is also a part of `Dockerfile.multiarch`. otherwise we have to set an extra variable for windows dockerfile too.

Makefile
jingxu97: In order to use buildx to build windows image on a linux machine, we cannot include any RUN command in the Dockerfile, so I think this limitation means we cannot build binary as a step in Dockerfile.

Makefile
jingxu97: Since Dockerfile.multiarch contains a 'RUN' command, we cannot use it to build windows image on a linux machine with buildx.

pohly: So we can build Windows binaries on the Linux host, but not inside a Docker container? Or does building for Windows need a Windows host?

pohly:
In https://github.com/kubernetes-csi/node-driver-registrar/pull/82/files/3e9e449f9f279c0c6e0859c1bce7c9b4b0007141#r425692239, we were discussing how to unify handling of Windows and Linux images. @jingxu97 pointed out that "Since Dockerfile.multiarch contains a 'RUN' command, we cannot use it to build windows image on a linux machine with buildx.". I was then wondering what we *can* use.

I tried it: we can and do cross-compile Windows binaries on a linux/amd64 host, just as we do for all other currently supported platforms (https://github.com/kubernetes-csi/csi-release-tools/blob/9084fecb84ddb47af8360d9ebe4f48a6b7a65c1d/prow.sh#L88).

As we are stuck with cross-compilation (correct, @jingxu97?), I am not sure whether we need to use "docker buildx". While in theory it may be nicer (building for multiple platforms with one instruction), we are not taking advantage of that and may as well do something like https://github.com/MarcusWichelmann/PrometheusSolarExporter/blob/3f2ee21b071af79a741cd1fefc02cba74c9bf166/.github/workflows/build.yml#L48-L65

@vitt-bagal, agreed that this could work?

However, one open question is whether can use the same base image or need something special for Windows. Is "gcr.io/distroless/static:latest" just for Linux and we have to use 
"mcr.microsoft.com/windows/nanoserver:1809" as base for Windows?

pohly:
I tried building multi-arch images with a local Docker registry, but I am running into https://github.com/docker/cli/issues/2396.

vitt-bagal:
> I tried building multi-arch images with a local Docker registry, but I am running into docker/cli#2396.

@pohly Even i also got the same issue with local docker registry. You can try with dockerhub.

namrata-ibm:
> As we are stuck with cross-compilation (correct, @jingxu97?), I am not sure whether we need to use "docker buildx". While in theory it may be nicer (building for multiple platforms with one instruction), we are not taking advantage of that and may as well do something like https://github.com/MarcusWichelmann/PrometheusSolarExporter/blob/3f2ee21b071af79a741cd1fefc02cba74c9bf166/.github/workflows/build.yml#L48-L65
> 
> @vitt-bagal, agreed that this could work?
> 
@pohly Docker buildx adds correct manifest for mutiarch images as we are using distroless image in Dockerfile. For more reference: https://github.com/bazelbuild/rules_docker/issues/300#issuecomment-574947315  

> However, one open question is whether can use the same base image or need something special for Windows. Is "gcr.io/distroless/static:latest" just for Linux and we have to use
> "mcr.microsoft.com/windows/nanoserver:1809" as base for Windows? 

Distroless seems to be for Linux only: https://github.com/GoogleContainerTools/distroless/tree/master/base#image-contents

@pohly how about testing Linux and Windows through different commands for now, considering various open issues?

Makefile
```diff
@@ -12,7 +12,46 @@
...
 include release-tools/build.make
+
+# This target builds multiarch images using Moby BuildKit builder toolkit.
+# Docker Buildx is included in Docker 19.03 and needs DOCKER_CLI_EXPERIMENTAL enabled to run corresponding commands.
+# Currently amd, s390x and Windows manifest is pushed for canary, release branch and released tags.
+# Images generated from Prow build are pushed to staging area on gcr
+push-multiarch-%:
+	make BUILD_PLATFORMS="windows amd64 .exe"
+	set -ex; \
+	gcloud auth configure-docker; \
```
msau42, line 29: This is going to require gcloud? Can we put this in the cloudbuild.yml instead?

Makefile
msau42: Any way we can make the list of linux platforms as a variable instead of hard coding the command for each platform?

Unfortunately I don't think there's much we can do to generalize windows atm.

Makefile
vitt-bagal: Earlier gcloud command was a part of cloudbuild only but later we have moved it to make target in order to keep cloudbuild simple and minimal.

Makefile
vitt-bagal: Currently with the help of this PR we are just testing that whether push-multiarch (keeping cloudbuild minimal) can be used to correctly tag multiarch images.
Once all these changes are working fine as expected then we can modify docker buildx command as `docker buildx build --push -t $(IMAGE_NAME):$$tag --platform=linux/amd64,linux/s390x -f ...` option which will have single docker buildx command for linux (with multiaple arch)

Makefile
msau42: I think it would be preferable to keep gcloud command outside of this otherwise anybody that wants to build a multiarch image is going to require gcloud to be installed.

If you still want put it in the makefile to reduce the commands in cloudbuild, then I would maybe wrap "push-multiarch" with a "cloudbuild-push-multiarch"

msau42:
For making progress on testing, this lgtm. My comments about making this more extensible can be addressed as a followup.
/lgtm
/approve

Makefile
pohly: I agree with @msau42 that this belongs into cloudbuild.yml. `push-multiarch-%` should be independent of the registry. `cloudbuild.yml` is where we prepare for pushing to gcr.

vitt-bagal:
@msau42 @pohly Latest image push [job](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/post-node-driver-registrar-push-images/1263528561675866113) failed. I could not see the build logs. Could you please help me with the build logs ?

pohly:
```
 *** No rule to make target 'push-multiarch REV=v$(echo v20200521-v1.3.0-34-g4dbc271e | cut -f3- -d 'v') REGISTRY_NAME=gcr.io/k8s-staging-csi'.  Stop.
```

The problem is the usage of "make" as entry point for the build. Variable expansion and splitting at spaces don't work when doing that. Obvious in retrospect, right? :stuck_out_tongue_closed_eyes: 

This might fix it (untested):
```
diff --git a/Makefile b/Makefile
index 32517b63..b8785ef8 100644
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,6 @@ include release-tools/build.make
 push-multiarch-%:
 	make BUILD_PLATFORMS="windows amd64 .exe"
 	set -ex; \
-	gcloud auth configure-docker; \
 	DOCKER_CLI_EXPERIMENTAL=enabled; \
 	export DOCKER_CLI_EXPERIMENTAL; \
 	docker buildx create --use --name multiarchimage-buildertest; \
diff --git a/cloudbuild.yaml b/cloudbuild.yaml
index e0d069df..00ab0287 100644
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -8,13 +8,17 @@ options:
   substitution_option: ALLOW_LOOSE
 steps:
   - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20200421-a2bf5f8'
-    entrypoint: make
+    entrypoint: sh
     env:
     - PULL_BASE_REF=$_PULL_BASE_REF
     - HOME=/root
     args:
-    # Extract tag-n-hash value from _GIT_TAG(form vYYYYMMDD-tag-n-hash) for REV value
-    - push-multiarch REV=v$(echo $_GIT_TAG | cut -f3- -d 'v') REGISTRY_NAME=gcr.io/k8s-staging-csi
+    - -e
+    - -c
+    - |
+      gcloud auth configure-docker
+      # Extract tag-n-hash value from _GIT_TAG(form vYYYYMMDD-tag-n-hash) for REV value
+      make push-multiarch REV=v$(echo $_GIT_TAG | cut -f3- -d 'v') REGISTRY_NAME=gcr.io/k8s-staging-csi
 substitutions:
   # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
   # can be used as a substitution
```

PR #84, opened 2020-05-25
multiarch image pushing

pohly:
**What type of PR is this?**
/kind feature

The previous PR (https://github.com/kubernetes-csi/node-driver-registrar/pull/82) had an issue (make invocation) and several opens. With the changes in this PR, `push-multiarch` integrates better with the rest of the build rules by removing duplication:
- the same Go environment is used for multiarch and for testing
- BUILD_PLATFORMS determines what is getting build instead
  of hard-coding that in the make target
- no separate `Dockerfile.multiarch`

Related-to https://github.com/kubernetes-csi/csi-release-tools/issues/86

The result of `make push-multiarch PULL_BASE_REF=master REGISTRY_NAME=pohly BUILD_PLATFORMS="linux amd64; windows amd64 .exe; linux ppc64le -ppc64le; linux s390x -s390x; linux arm64 -arm64"` can be found here: https://hub.docker.com/repository/docker/pohly/csi-node-driver-registrar [Update 2020-05-26: added arm64]

```release-note
The Dockerfile must accept a "binary" build argument.
```


pohly:
> the same Go environment is used for multiarch and for testing

I need to take this back a bit: this is only true for local builds and Prow, but not for cloud build. There https://github.com/kubernetes-csi/node-driver-registrar/blob/cae4fbec4fd7e6f8a174a5c6091be01bd1743c9c/cloudbuild.yaml#L10 determines which version of Go is used.

I think we should use the same approach as for Prow and install the desired Go version if it isn't available already.

pohly:
> I think we should use the same approach as for Prow and install the desired Go version if it isn't available already.

I've added that.

Makefile
```diff
@@ -19,39 +19,65 @@ all: build
...
-	gcloud auth configure-docker; \
 	DOCKER_CLI_EXPERIMENTAL=enabled; \
 	export DOCKER_CLI_EXPERIMENTAL; \
-	docker buildx create --use --name multiarchimage-buildertest; \
+	docker buildx create $(DOCKER_BUILDX_CREATE_ARGS) --use --name multiarchimage-buildertest; \
+	trap "docker buildx rm multiarchimage-buildertest" EXIT; \
+	dockerfile_linux=$$(if [ -e ./cmd/$*/Dockerfile ]; then echo ./cmd/$*/Dockerfile; else echo Dockerfile; fi); \
+	dockerfile_windows=$$(if [ -e ./cmd/$*/Dockerfile.Windows ]; then echo ./cmd/$*/Dockerfile.Windows; else echo Dockerfile.Windows; fi); \
+	build_platforms='$(BUILD_PLATFORMS)'; \
+	if ! [ "$$build_platforms" ]; then build_platforms="linux amd64"; fi; \
```
msau42, line 46: Can we set the default in L45?

Makefile
```diff
@@ -19,39 +19,65 @@ all: build
...
+			release_canary_tag=$$(echo $(PULL_BASE_REF) | cut -f2 -d '-')-canary; \
+			pushMultiArch $$release_canary_tag; \
 	elif docker pull $(IMAGE_NAME):$(PULL_BASE_REF) 2>&1 | tee /dev/stderr | grep -q "manifest for $(IMAGE_NAME):$(PULL_BASE_REF) not found"; then \
-                       : "creating release image"; \
-                       pushMultiArch $(PULL_BASE_REF); \
+		       : "creating release image"; \
+		       pushMultiArch $(PULL_BASE_REF); \
 	else \
-                       : "release image $(IMAGE_NAME):$(PULL_BASE_REF) already exists, skipping push"; \
+		       : "release image $(IMAGE_NAME):$(PULL_BASE_REF) already exists, skipping push"; \
```
msau42, line 73: should this be silently ignored, or return an error?

cloudbuild.yaml
```diff
@@ -7,14 +7,18 @@ timeout: 1200s
...
+  # the desired version of Go (currently defined in release-tools/travis.yml),
+  # but that just speeds up the build and is not required.
   - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20200421-a2bf5f8'
-    entrypoint: make
+    entrypoint: bash
     env:
-    - PULL_BASE_REF=$_PULL_BASE_REF
+    - GIT_TAG=${_GIT_TAG}
+    - PULL_BASE_REF=${_PULL_BASE_REF}
+    - STAGING_PROJECT=${_STAGING_PROJECT}
```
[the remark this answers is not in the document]
pohly: You are right, setting `REGISTRY_NAME` directly is cleaner. I'm not even sure why we have `_STAGING_PROJECT` under substitutions.

@vitt-bagal: do you remember where that came from? Isn't that always `k8s-staging-csi`?

Makefile
pohly: I haven't thought about this much. The whole if/else sequence here was copied originally from the corresponding "push" target and I just kept it. The original motivation for the check was that we directly push to the final destination, and those images should never be rebuilt because someone may already have pulled them. Here the situation is a bit different.

Do we expect that the cloud image build job will run more than once per tag? If no, then we can remove the "does the image exist" check because the image should never exist already. If yes, perhaps the build job was triggered intentionally again and the goal was to rebuild the image.

I'm leaning towards removing the check.

cloudbuild.yaml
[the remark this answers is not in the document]
vitt-bagal: > @vitt-bagal: do you remember where that came from? Isn't that always k8s-staging-csi?

It came from this [commit](https://github.com/kubernetes-csi/node-driver-registrar/commit/a44583857c746a1de0cc26a74e2ab1446adf3b65). I think this can be removed safely if we set `REGISTRY_NAME` directly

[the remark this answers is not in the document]
pohly: If we can move `cloudbuild.yaml` into `release-tools` and just symlink to it from the root directory of a project (similar to `.travis.yml -> release-tools/travis.yml`), then it may make sense to have the substitution because different projects might have different staging projects.

Makefile
pohly: `$(BUILD_PLATFORMS)` is expanded by make, so the usual shell mechanism (`${var:-default}`) does not work.

I've combined the two lines into one with:
```
if [ '$(BUILD_PLATFORMS)' ]; then build_platforms='$(BUILD_PLATFORMS)'; else build_platforms="linux amd64"; fi;
```

Makefile
msau42: I think we should never repush an existing tag (except for canary), even if there was some error. We would have to resolve that scenario by incrementing the tag.

pohly: Okay, it's an error now.

pohly:
@msau42: please take another look

PR #95, opened 2020-07-23
Add flag to configure the registration socket path

### LINKED ISSUE #94: FeatureRequest: Add support for configuring the registration socket path
As per the current code we expect `/var/lib/kubelet/plugins_registry/` to be mounted as `/registration`. This works well when `node-driver-registrar` runs as sidecar container or as a pod. We have a usecase to run node-driver-registrar as systemd process to support csi driver for specific set of worker nodes (https://github.com/aws/containers-roadmap/issues/826). Therefore, instead of hardcoding `/registartion` would be helpful if we can read this value from a parameter. If the new parameter is empty then it can default to `/registration`

Would like to get community feedback on whether this is acceptable or is the node-driver-registrar designed to run only as side-car container?. if we decide to add new parameter then naming might be a difficult part I think, given that we already have parameter named `kubelet-registrataion-path` which points to csi socket path.

prithviramesh:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature

> /kind flake

adds support to dynamically configure the registration socket path via a `plugin-registration-path` flag 

Fixes #94 

```release-note
the addition of the --plugin-registration-path flag allows users to dynamically configure the registration path; this flag defaults to the currently hardcoded value of `/registration` so actions can be taken to consume this change but is not required
```

### DIFF
--- a/cmd/csi-node-driver-registrar/main.go
+++ b/cmd/csi-node-driver-registrar/main.go
@@ -46,6 +46,7 @@ const (
 var (
 	connectionTimeout       = flag.Duration("connection-timeout", 0, "The --connection-timeout flag is deprecated")
 	csiAddress              = flag.String("csi-address", "/run/csi/socket", "Path of the CSI driver socket that the node-driver-registrar will connect to.")
+	pluginRegistrationPath  = flag.String("plugin-registration-path", "/registration", "Path to Kubernetes plugin registration directory.")
 	kubeletRegistrationPath = flag.String("kubelet-registration-path", "", "Path of the CSI driver socket on the Kubernetes host machine.")
 	showVersion             = flag.Bool("version", false, "Show version.")
 	version                 = "unknown"
--- a/cmd/csi-node-driver-registrar/node_register.go
+++ b/cmd/csi-node-driver-registrar/node_register.go
@@ -75,7 +75,7 @@ func nodeRegister(
 }
 
 func buildSocketPath(csiDriverName string) string {
-	return fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+	return fmt.Sprintf("%s/%s-reg.sock", *pluginRegistrationPath, csiDriverName)
 }
 
 func removeRegSocket(csiDriverName string) {


cmd/csi-node-driver-registrar/main.go
```diff
@@ -46,6 +46,7 @@ const (
 var (
 	connectionTimeout       = flag.Duration("connection-timeout", 0, "The --connection-timeout flag is deprecated")
 	csiAddress              = flag.String("csi-address", "/run/csi/socket", "Path of the CSI driver socket that the node-driver-registrar will connect to.")
+	pluginRegistrationPath  = flag.String("plugin-registration-path", "/registration", "Path of the registration socket on the Kubernetes host machine.")
```
jsafrane, line 49: It's not "not "path of the registration socket" - it directory name. And it's not anything on Kubernetes host machine, but path where the plugin registration directory is exposed in the registrar container,

prithviramesh: ah - we're using this not in a sidecar container, but running the registrar directly as a `systemd` process on a host machine (Fargate task)

prithviramesh: I can update the description as you guys see fit.

jsafrane: What about "Path to Kubernetes plugin registration directory"?

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -36,7 +36,7 @@ func nodeRegister(
 	// as gRPC server which replies to registration requests initiated by kubelet's
 	// pluginswatcher infrastructure. Node labeling is done by kubelet's csi code.
 	registrar := newRegistrationServer(csiDriverName, *kubeletRegistrationPath, supportedVersions)
-	socketPath := fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+	socketPath := fmt.Sprintf("%s/%s-reg.sock", *pluginRegistrationPath, csiDriverName)
```
jsafrane, line 39: maybe `filepath.Join`?

cmd/csi-node-driver-registrar/main.go
jsafrane: @prithviramesh, can you please update the description?

prithviramesh: I can, what you suggested SGTM. Will push up another rev today

jsafrane:
@prithviramesh, the code around sockets got reworked a bit, see https://github.com/kubernetes-csi/node-driver-registrar/pull/61
Can you please rebase?

prithviramesh:
@jsafrane I went ahead and rebased - sorry about the delay here.

prithviramesh:
Looks like all the tests passed post rebase, @jsafrane let me know if we need anything else to get this merged in.
Thanks for your feedback so far!

PR #96, opened 2020-07-24
bump pluginregistration api version to v1

gnufied:
We have been using v1alpha1 version for far too long. 

cc @msau42 @jsafrane 

```release-note
Use v1 API for plugin registration with kubelet
```


humblec:
@gnufied can you move to `v0.19.0-rc.2` as part of this PR?

gnufied:
@humblec why? I would leave it at a released version, there is literally no change in API, the main thing is we want to stop using v1alpha1

jsafrane:
Is this a breaking change? The registrar stops working in Kubernetes v1.13. It's OK to break it, however, we should probably bump the registrar to v2.

jsafrane:
Filed https://github.com/kubernetes-csi/node-driver-registrar/issues/97

humblec:
> @humblec why? I would leave it at a released version, there is literally no change in API, the main thing is we want to stop using v1alpha1

sure.. :).. I thought we are going to update this anyway to 1.19 soon. Thats why I mentioned we could try now itself :)

PR #99, opened 2020-08-10
HTTP server for liveness probe, check registration socket exists

### LINKED ISSUE #57: Ensure reg.sock exists while the registrar is running
While the registrar is running, the kubelet registration socket it installs could be removed erroneously by a different process (for example if two instances of the same driver are running in the cluster). The registrar should continuously ensure that the socket exists during its lifetime (and subsequently removed before it terminates).

/cc @davidz627 @msau42

chrishenzie:
**What type of PR is this?**

/kind feature

Adds a liveness probe server that monitors the existence of the registration socket. Included are refactors to reduce code duplication checking for socket exists. Note that there are behavioral differences in how linux and windows systems check for sockets, see the bug in comments below.

Fixes #57 

```release-note
Adds a new flag --health-port. This is the port of the health check server for the node-driver-registrar, which checks if the registration socket exists. A value <= 0 disables the server. Server is disabled by default.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,12 @@ There are two UNIX domain sockets used by the node-driver-registrar:
   `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock). Note this is NOT
   the path to the registration socket.
 
+### Optional arguments
+
+* `--health-port`: This is the port of the health check server for the node-driver-registrar,
+  which checks if the registration socket exists. A value <= 0 disables the server.
+  Server is disabled by default.
+
 ### Required permissions
 
 The node-driver-registrar does not interact with the Kubernetes API, so no RBAC
@@ -76,11 +82,21 @@ the actual driver's name.
           args:
             - "--csi-address=/csi/csi.sock"
             - "--kubelet-registration-path=/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock"
+            - "--health-port=9809"
           volumeMounts:
             - name: plugin-dir
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
+          ports:
+            - containerPort: 9809
+              name: healthz
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: healthz
+            initialDelaySeconds: 5
+            timeoutSeconds: 5
       volumes:
         - name: registration-dir
           hostPath:
--- a/cmd/csi-node-driver-registrar/main.go
+++ b/cmd/csi-node-driver-registrar/main.go
@@ -47,6 +47,7 @@ var (
 	connectionTimeout       = flag.Duration("connection-timeout", 0, "The --connection-timeout flag is deprecated")
 	csiAddress              = flag.String("csi-address", "/run/csi/socket", "Path of the CSI driver socket that the node-driver-registrar will connect to.")
 	kubeletRegistrationPath = flag.String("kubelet-registration-path", "", "Path of the CSI driver socket on the Kubernetes host machine.")
+	healthzPort             = flag.Int("health-port", 0, "TCP port for healthz requests. Set to 0 to disable the healthz server.")
 	showVersion             = flag.Bool("version", false, "Show version.")
 	version                 = "unknown"
 
--- a/cmd/csi-node-driver-registrar/node_register.go
+++ b/cmd/csi-node-driver-registrar/node_register.go
@@ -19,9 +19,11 @@ package main
 import (
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"runtime"
+	"strconv"
 	"syscall"
 
 	"google.golang.org/grpc"
@@ -64,6 +66,7 @@ func nodeRegister(
 	// Registers kubelet plugin watcher api.
 	registerapi.RegisterRegistrationServer(grpcServer, registrar)
 
+	go healthzServer(socketPath, *healthzPort)
 	go removeRegSocket(csiDriverName)
 	// Starts service
 	if err := grpcServer.Serve(lis); err != nil {
@@ -76,6 +79,33 @@ func nodeRegister(
 
 func buildSocketPath(csiDriverName string) string {
 	return fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
+}
+
+func healthzServer(socketPath string, port int) {
+	if port <= 0 {
+		klog.Infof("Skipping healthz server because port set to: %v", port)
+		return
+	}
+	klog.Infof("Starting healthz server on port: %v\n", port)
+
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, req *http.Request) {
+		socketExists, err := util.DoesSocketExist(socketPath)
+		if err == nil && socketExists {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`ok`))
+			klog.V(5).Infof("health check succeeded")
+		} else if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(err.Error()))
+			klog.Errorf("health check failed: %+v", err)
+		} else if !socketExists {
+			w.WriteHeader(http.StatusNotFound)
+			w.Write([]byte("registration socket does not exist"))
+			klog.Errorf("health check failed, registration socket does not exist")
+		}
+	})
+
+	klog.Fatal(http.ListenAndServe(":"+strconv.Itoa(port), nil))
 }
 
 func removeRegSocket(csiDriverName string) {
--- a/pkg/util/util_linux.go
+++ b/pkg/util/util_linux.go
@@ -30,15 +30,25 @@ func Umask(mask int) (int, error) {
 }
 
 func CleanupSocketFile(socketPath string) error {
-	fi, err := os.Stat(socketPath)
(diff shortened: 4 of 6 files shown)


chrishenzie:
Replying to test CLA.

pkg/util/util_windows.go
```diff
@@ -29,19 +29,27 @@ func Umask(mask int) (int, error) {
...
-	if _, err := os.Lstat(socketPath); err != nil {
-		// If the file does not exist, then the cleanup can be considered successful.
-		if os.IsNotExist(err) {
-			return nil
+	socketExists, err := DoesSocketExist(socketPath)
+	if err != nil {
+		return err
+	}
+	if socketExists {
+		// TODO: Until the bug - https://github.com/golang/go/issues/33357 is fixed, os.stat wouldn't return the
```
msau42, line 37: I think this comment fits better under `DoesSocketExist` to describe why we're not checking the file mode.

cmd/csi-node-driver-registrar/main.go
```diff
@@ -47,6 +47,7 @@ var (
 	connectionTimeout       = flag.Duration("connection-timeout", 0, "The --connection-timeout flag is deprecated")
 	csiAddress              = flag.String("csi-address", "/run/csi/socket", "Path of the CSI driver socket that the node-driver-registrar will connect to.")
 	kubeletRegistrationPath = flag.String("kubelet-registration-path", "", "Path of the CSI driver socket on the Kubernetes host machine.")
+	livenessProbePort       = flag.Int("liveness-probe-port", 9090, "HTTP port for liveness probe server.")
```
jsafrane, line 50: CSI drivers on nodes often run with `hostNetwork: true`, which means that the liveness probe port must be open on the node. It becomes increasingly more complicated to make sure the port does not clash with any other port that the node needs.

It should be at least possible to turn the probe off (`--liveness-probe-port=0`?)

jsafrane:
Now that I think about this, can the registrar just have `inotify` on the socket? If another driver registrar starts on the same node (e.g. someone deploys a second DaemonSet for the same driver), the first one should detect that its sockets are deleted and exit. With liveness probe, it is likely to miss the time when the socket is deleted and the old registrar will continue running even if there is another registrar serving the socket.

chrishenzie:
@jsafrane Just clarifying the approach, you would prefer the liveness probe is removed and instead `os.Exit()` when the registration socket is deleted by `util.CleanupSocketFile()`? It looks like [this fsnotify library](https://github.com/fsnotify/fsnotify) may help, am I free to add this dependency or are there other approaches to consider?

msau42:
> Now that I think about this, can the registrar just have `inotify` on the socket? If another driver registrar starts on the same node (e.g. someone deploys a second DaemonSet for the same driver), the first one should detect that its sockets are deleted and exit. With liveness probe, it is likely to miss the time when the socket is deleted and the old registrar will continue running even if there is another registrar serving the socket.

What prevents the restarted driver-registrar from coming up again, recreating the socket, and then ping-pong back and forth forever?

cmd/csi-node-driver-registrar/main.go
msau42: Maybe don't default the port, and if unset, then don't enable this?

jsafrane:
> What prevents the restarted driver-registrar from coming up again, recreating the socket, and then ping-pong back and forth forever?

Nothing. But letting two registrar pods listen on the same socket is IMO even worse.

msau42:
> letting two registrar pods listen on the same socket

I don't think this is what's happening. The 2nd registrar will come up, delete the registration socket, and listen on the new one. The 1st driver-registrar doesn't get any requests. I believe @chrishenzie verified this by checking the logs of both. 

I was surprised that the 1st registrar didn't error out when the socket file it was listening on got removed.

msau42:
The purpose of this PR is to make it so that: 1) if 2 of the same driver is running on the same node, node operations are still functional. and 2) when one of the duplicate driver instances is removed and the socket gets deleted, it doesn't cause the driver to be unregistered forever.

cmd/csi-node-driver-registrar/main.go
jsafrane: That could work. Still, I wish there was an exec probe instead.

jsafrane:
> The 2nd registrar will come up, delete the registration socket, and listen on the new one. The 1st driver-registrar doesn't get any requests

If that's the case, then I withdraw my objections. It's quite surprising though.

Still, please make the port optional.

cmd/csi-node-driver-registrar/main.go
msau42: Can you clarify what you mean by exec probe?

README.md
```diff
@@ -50,6 +50,11 @@ There are two UNIX domain sockets used by the node-driver-registrar:
   `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock). Note this is NOT
   the path to the registration socket.
 
+### Optional arguments
+
+* `--health-port`: This is the port of the health check server for the node-driver-registrar.
```
msau42, line 55: Can you briefly describe what the health check is doing here?

README.md
```diff
@@ -50,6 +50,11 @@ There are two UNIX domain sockets used by the node-driver-registrar:
...
+  A value of 0 will disable the server.
```
msau42, line 56: Also mention by default it's disabled

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -78,6 +81,33 @@ func buildSocketPath(csiDriverName string) string {
 	return fmt.Sprintf("/registration/%s-reg.sock", csiDriverName)
 }
 
+func healthzServer(socketPath string, port int) {
+	if port == 0 {
```
msau42, line 85: check for negative values too

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -78,6 +81,33 @@ func buildSocketPath(csiDriverName string) string {
...
+		if err == nil && socketExists {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte(`ok`))
+			klog.V(5).Infof("health check succeeded")
+		} else if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte(err.Error()))
+			klog.Errorf("health check failed: %+v", err)
+		} else if !socketExists {
+			w.WriteHeader(http.StatusInternalServerError)
```
msau42, line 102: Do we want to distinguish error codes and have this one be `StatusNotFound`?

pkg/util/util_linux.go
```diff
@@ -30,15 +30,25 @@ func Umask(mask int) (int, error) {
...
 		if err := os.Remove(socketPath); err != nil {
 			return fmt.Errorf("failed to remove stale socket %s with error: %+v", socketPath, err)
 		}
 	}
+	return nil
+}
+
+func DoesSocketExist(socketPath string) (bool, error) {
+	fi, err := os.Stat(socketPath)
+	if err == nil && (fi.Mode()&os.ModeSocket) != 0 {
```
msau42, line 47: This isn't your change (and maybe we can think about this in a followup), but I wonder if it's worth it to log a warning or return an error if the file is not a socket instead of returning false.  I wouldn't expect creating a listening socket would work if the file already existed and it's not a socket file.

README.md
```diff
@@ -76,11 +81,19 @@ the actual driver's name.
...
+            - "--health-port=9809"
           volumeMounts:
             - name: plugin-dir
               mountPath: /csi
             - name: registration-dir
               mountPath: /registration
+          ports:
+            - containerPort: 9809
+              name: healthz
+          livenessProbe:
```
msau42, line 93: Should we set an initialDelay to give some time to create the socket? And should we set timeoutSeconds to give some grace period in case the system is very busy?

The other defaults seem reasonable to me

msau42:
Can you also add a release note in the PR description that mentions the new flag added and what it does?

README.md
```diff
@@ -76,11 +81,19 @@ the actual driver's name.
...
+            httpGet:
```
msau42, line 94: Can you also test this out in a driver Pod where `hostNetwork: true`? I'm not sure if anything special needs to be done to configure the livenessProbe in that scenario

README.md
```diff
@@ -76,11 +81,19 @@ the actual driver's name.
           args:
```
msau42, line 81: @jsafrane you missed updating the image registry here :-P

cmd/csi-node-driver-registrar/main.go
chrishenzie: I believe they mean doing something like this:
```
livenessProbe:
  exec:
    command:
      - stat
      - /registration/io.kubernetes.storage.mock-reg.sock
```

pkg/util/util_linux.go
chrishenzie: Yeah good point, if we dropped something else there with the same name the registrar wouldn't be able to create a socket.

README.md
chrishenzie: Just tested on a GCE node, didn't encounter any issues, was able to see the health checks pass (and fail when I removed the registration port).

README.md
chrishenzie: I can tackle this in a separate PR.

msau42:
/lgtm
/approve

Thanks for adding this!

PR #122, opened 2020-11-19
Bump csi-lib-utils to v0.9.0

jsafrane:
/kind cleanup

To get rid of klogv1.

It was necessary to create a dummy `CSIMetricsManager`, which does not report metrics anywhere. It may report them if we wanted, but that's for another PR. And it would need to share http endpoint with `/healthz`.

```release-note
NONE
```


cmd/csi-node-driver-registrar/main.go
```diff
@@ -115,12 +116,15 @@ func main() {
 		klog.Warning("--connection-timeout is deprecated and will have no effect")
 	}
 
+	// Unused metrics manager, necessary for connection.Connect below
+	cmm := metrics.NewCSIMetricsManagerForSidecar("")
```
pohly, line 120: I suppose it wasn't made optional because all sidecars are expected to actually server metrics data.

pohly (approved):
/lgtm

PR #124, opened 2020-11-26
add --timeout for gRPC calls

pohly:
**What type of PR is this?**
/kind feature

The one second timeout is based on the assumption that once the
connection is established, all further communication is going to be
fast.

That assumption doesn't hold in some special circumstances:
- interactively debugging the sidecar
- proxying the driver operation to some remote implementation

While this is a bit obscure, the flag is essential for these
cases. Adding the flag also makes the node-driver-registrar consistent
with other sidecars.

I have a prototype working with the mock driver embedded inside the e2e.test binary (the "proxying operation" case). I'll share more details after completing that work. In the meantime it would be good to get this PR included in the next release for Kubernetes 1.20.

```release-note
added --timeout parameter for the unlikely case that a driver's GetPluginInfo call is slower than one second
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ There are two UNIX domain sockets used by the node-driver-registrar:
   node-driver-registrar, which checks if the registration socket exists. A value <= 0 disables
   the server. Server is disabled by default.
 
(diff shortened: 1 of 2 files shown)


pohly:
/assign @jsafrane 

For approval.

pohly:
/assign @msau42 

Looks like Jan is busy today, let's see whether Michelle has time for the final approval...

Issue #104, opened 2020-08-17
Return an error when registration socket file exists but is not a socket


msau42:
Followup to https://github.com/kubernetes-csi/node-driver-registrar/pull/99/files#r469647510
/help
/good-first-issue

Issue #93, opened 2020-07-22
Upgrade grpc-go dependency to allow drivers to be written in rust / typescript


entropitor:
Due to a bug in grpc-go, drivers in typescript / rust weren't working. This has been fixed so now this package needs to update its dependency on grpc-go soon (https://github.com/grpc/grpc-go/issues/2628)

msau42:
From the issue it looks like this fix is going to be in grpc 1.31?
/help
/good-first-issue

entropitor:
@Jiawei0227 Do we still need https://github.com/kubernetes/kubernetes/issues/93320 to be merged as well or should this be enough to start writing CSI's in other languages without any hacks?

PR #126, opened 2020-12-08
Upgrade grpc-go dependency and check if file in SocketPath is actually a socket

Jiawei0227:
**What type of PR is this?**
/kind bug

1. Return error when registration socket file exists but is not a socket
2. Upgrade grpc-go dependency to 1.31

Fixes # https://github.com/kubernetes-csi/node-driver-registrar/issues/104
Fixes # https://github.com/kubernetes-csi/node-driver-registrar/issues/93

```release-note
Return error when file in SocketPath exists but is not a socket; Upgrade grpc-go dependency to 1.31
```

/assign @chrishenzie 
/cc @msau42


pkg/util/util_linux.go
```diff
@@ -44,8 +45,11 @@ func CleanupSocketFile(socketPath string) error {
 
 func DoesSocketExist(socketPath string) (bool, error) {
 	fi, err := os.Stat(socketPath)
-	if err == nil && (fi.Mode()&os.ModeSocket) != 0 {
-		return true, nil
+	if err == nil {
+		if isSocket := (fi.Mode()&os.ModeSocket != 0); isSocket {
+			return true, nil
+		}
+		klog.Warningf("file exists in socketPath %s but it's not a socket.: %+v", socketPath, fi)
```
chrishenzie, line 52: Does this fix https://github.com/kubernetes-csi/node-driver-registrar/issues/104? It seems like this will return `false` and `nil` for the error if it is not a socket.

chrishenzie:
Do we have a unit test that asserts we should return an error if the socket path exists but is not a socket file?

pkg/util/util_linux.go
Jiawei0227: Yeah good point... I initially return error but then wonder if it will break things so I change it to warning. I think an Error here seems okay to me? It's just we will have to change the CleanUpSocket function as well.

chrishenzie: I vaguely recall the socket checking logic differs between linux and Windows. It might be worth double-checking.

Jiawei0227: There is already a TestSocketRegularFile which I modified a little bit that test this behavior.
The limitation is that we will have different behaviors on linux and windows... Because according https://github.com/golang/go/issues/33357, windows socket file are not returning the right mode somehow... So the test will behavior a little bit different based on the two platforms.

PR #129, opened 2020-12-14
add support to configure logging format

### LINKED ISSUE #128: support JSON log format for node-driver-registrar
- Provide a flag to enable JSON log formatting. 

We do this today for CSI driver implementations but the same can't be configured for `node-driver-registrar`. This can be done by adding `log-format-json=true` flag to the node-driver-registrar and setting the log format based on that.

aramase:
**What type of PR is this?**

/kind feature

Adds configuration for json log formatter

Fixes #128 

```release-note
added --logging-format to set log formatter (default: text)
```


Jiawei0227:
Hey @aramase , can you help to rebase and update the dependency to 0.20? Thanks a lot!

aramase:
> Hey @aramase , can you help to rebase and update the dependency to 0.20? Thanks a lot!

@Jiawei0227 for sure!

Jiawei0227:
/lgtm
Thanks for the contribution! 
/assign @msau42 for the final approval.

cmd/csi-node-driver-registrar/main.go
```diff
@@ -53,6 +54,7 @@ var (
 	healthzPort             = flag.Int("health-port", 0, "(deprecated) TCP port for healthz requests. Set to 0 to disable the healthz server. Only one of `--health-port` and `--http-endpoint` can be set.")
 	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
 	showVersion             = flag.Bool("version", false, "Show version.")
+	logFormatJSON           = flag.Bool("log-format-json", false, "set log formatter to json")
```
msau42, line 57: It looks like component base already supports a log format flag, so I'm not sure we need to explicitly add one here: 

https://github.com/kubernetes/component-base/blob/master/logs/options.go#L89

I think it would be better if we switch to component-base log pkg instead of using klog directly.

aramase: sure, let me do that.

aramase: Updated, PTAL!

cmd/csi-node-driver-registrar/main.go
```diff
@@ -18,13 +18,15 @@ package main
 
 import (
 	"context"
-	"flag"
+	goflag "flag"
```
Jiawei0227, line 21: How about get rid of goflag and only use pflag?
I am wondering if something like this will just do the magic.
`flag "github.com/spf13/pflag"`

msau42:
Also cc @pohly @jsafrane due to previous issues we had initializing klog correctly.

cmd/csi-node-driver-registrar/main.go
aramase: So I did switch this to pflag, but we still need to add "flag" flagset to allow configuring klog flags

Jiawei0227: Sorry I am super unfamiliar to this, what goflag do we need to add to pflag? I thought in your change you already define all the flag in pflag?

aramase: So klog uses the default `flag.FlagSet` for defining the flags: https://github.com/kubernetes/klog/blob/master/klog.go#L420-L441. By doing `pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)`, we'll be importing all the flags defined for `flag` into `pflag`. 

If we don't add the `flag.FlagSet` to `pflag`, then setting verbosity and other klog flags will fail:
```
➜ ./bin/livenessprobe -v=2
unknown shorthand flag: 'v' in -v=2
```

Jiawei0227: Thanks a lot! This looks good to me!

cmd/csi-node-driver-registrar/main.go
```diff
@@ -99,9 +101,18 @@ func (e registrationServer) NotifyRegistrationStatus(ctx context.Context, status
...
-	flag.Set("logtostderr", "true")
-	flag.Parse()
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
+	logOptions := logs.NewOptions()
+	logOptions.AddFlags(pflag.CommandLine)
+
+	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
+	pflag.Set("logtostderr", "true")
```
pohly, line 111: Do we still need to set `logtostderr`? It might be the default in klog/v2 now.

If we can simplify it, now would be a good time.

pohly: The flag is now called `alsologtostderr` - I doubt that this `pflag.Set` does anything :grin:

pohly (commented):
Careful!

Switching from `flag` to `pflag` is a breaking change because single-dash command line parameters no longer work.

This used to be valid:
```
csi-node-driver-registrar -v=5 -csi-address=/tmp/csi.sock -kubelet-registration-path=/tmp
```

With this PR, it fails:
```
unknown shorthand flag: 'c' in -csi-address=/tmp/csi.sock
```

IMHO this is too big of a change and should better be avoided.

Incidentally, I'm also working on JSON output in PMEM-CSI right now. We also used the normal `flag` there. To hook up `k8s.io/component-base/logs` to that, I used this:

helper package:
```
/*
Copyright 2020 Intel Coporation.

SPDX-License-Identifier: Apache-2.0
*/

package logger

import (
	"flag"

	utilerrors "k8s.io/apimachinery/pkg/util/errors"
	"k8s.io/component-base/logs"
)

func NewFlag() *Options {
	f := &Options{}
	flag.Var(f, "logging-format", "determines log output format, 'text' and 'json' are supported")
	return f
}

// Options is a wrapper around Options which makes
// it usable with flags.Var.
type Options struct {
	logs.Options
}

func (f *Options) Set(value string) error {
	f.LogFormat = value
	return utilerrors.NewAggregate(f.Validate())
}

func (f *Options) String() string {
	return f.LogFormat
}
```

main:
```
var logFormat   = logger.NewFlag()
...

func main() int {
	flag.Parse()
	logFormat.Apply()
```

It's less nice than using `pflag` because the available logging formats are not listed automatically, but it does the job.

/hold

@msau42 what do you think? Use `pflag` and do a new major release or stick with `flag`?

pohly:
FWIW, I'd like to see JSON output supported in all sidecars. Considering that, perhaps the necessary boiler plate code (like this helper package) should be in csi-lib-utils? We need to avoid the list of fixed formats in the lib, though, because a binary might add additional formats.

aramase:
> Incidentally, I'm also working on JSON output in PMEM-CSI right now. We also use the normal flag there. To hook up k8s.io/component-base/logs to that, I used this:

Thank you for the feedback. I'm have the exact same change in a different project using flags flagset similar to what you posted. I'm happy to update this PR based on the approach we want to take.

msau42:
I think switching to component-base logs is still desirable. There are some helpers with log deduplication and sanitization that seem useful to take advantage of. It's unfortunate that the flags are not compatible though, I was hoping we wouldn't have to define our own flags, but it doesn't seem like we can do that to retain backwards compatibility.

What do you think of this plan?
1. Add the wrapper now to retain backwards compatibility, but we deprecate support for the single-dash flags.
2. When we're ready for another major release, we remove this wrapper and switch directly to using component-base flags.

I think over time, we can adopt more things from component-base, so this issue will come up again, so it's good to standardize in the long term.

msau42:
We also discussed in today's standup meeting. This is a change that we want for all our sidecars, but it's going to be a bit late to add this because we're cutting releases today before everyone goes on holiday break. We'll have to target this change for our next round of releases.

aramase:
> We also discussed in today's standup meeting. This is a change that we want for all our sidecars, but it's going to be a bit late to add this because we're cutting releases today before everyone goes on holiday break. We'll have to target this change for our next round of releases.


Sounds good! As this change will not make it for today's release, should we still go with the 2 step plan? I can make the changes to add the wrapper for now.

pierluigilenoci:
@aramase could you please fix the conflict?

pierluigilenoci:
@aramase any feedback from the community?
Ref: https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/442#issuecomment-777588848

msau42:
Revisiting this again. Now that Kubernetes has introduced [structured logging](https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/), do we still want the changes here, or do we want to consider following what Kubernetes is doing?

aramase:
> Revisiting this again. Now that Kubernetes has introduced structured logging, do we still want the changes here, or do we want to consider following what Kubernetes is doing?

Even with structured logging we would need to support the json log format so I think we would still need this PR. I can help with using structured logging in a separate PR for the node-driver-registrar, liveness-probe and dust off this PR for review again. Let me know.

cmd/csi-node-driver-registrar/main.go
pohly: If I remember correctly, switching from flag to pflag is a breaking change because the single-dash options that were valid are no longer accepted.

pierluigilenoci:
@aramase any feedback from the community?
Ref: https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/442#issuecomment-777588848

aramase:
> @aramase any feedback from the community?
Ref: https://github.com/kubernetes-sigs/secrets-store-csi-driver/issues/442#issuecomment-777588848

@pierluigilenoci There has been some discussion on the PRs. I will let the maintainers of the node-driver-registrar and liveness probe decide how they want to proceed with supporting json log formatting. I don't have the time currently for rebasing the 2 PRs, so I'm going to close them and let someone else pick up the issue if there is interest in supporting the json log format.

/close

pohly:
Work in this area will be tracked via this project board: https://github.com/orgs/kubernetes-csi/projects/46

PR #133, opened 2021-01-28
feat(util): Allows building for darwin targets

thomastaylor312:
**What type of PR is this?**

/kind feature

This is a simple change that allows the registrar to be built for MacOS. All of the `unix` package being used works for Mac as well. The reason for enabling this is completely selfish I'll admit. We are adding [CSI Support](https://github.com/deislabs/krustlet/pull/462) to the Krustlet project, which supports Mac as well. As most CSI implementations don't support registering themselves with the Kubelet directly, we wanted to unblock any users to need to register their plugins with a Krustlet Kubelet. This change shouldn't affect normal container-based usage of the registrar

```release-note
NONE
```


msau42:
@pohly do you know if we can add this to our build job to make sure this compiles? Better yet, if we can release images with it?

mauriciopoppe:
Sorry for creating a rebase, I think that the first line should also be changed to [`//go:build !windows`](https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md), I think it was added by my editor on save

thomastaylor312:
Sweet, I'll fix that real quick

pkg/util/util_nix.go
```diff
@@ -1,5 +1,5 @@
-//go:build linux
-// +build linux
+//go:build !windows
```
msau42, line 1: This is the first time I've seen the "nix" suffix being used. Is there a general convention that the Kubernetes project is using to support darwin?

pohly: @thomastaylor312: can you explain where the `_nix.go` filename comes from? Is that being used elsewhere already?

thomastaylor312: We can just call it `unix` if you prefer. I was just trying to indicate that this file wasn't for just linux, but any *nix system. I don't have any sort of a strong preference. Also no skin off of my teeth if you don't want to support darwin targets. This was just a proposal of something to do and can be shot down. As you can see, there was no change to code, so the restriction to linux build targets only was artificial and not for any specific code reason insofar as I can tell

pohly: I (like @msau42 I guess) was just wondering. Personally I prefer "unix" over "nix". The latter might not be immediately recognizable as a placeholder for "unix or linux".

Kubernetes releases darwin/amd64 and darwin/arm64 client binaries, but not node binaries. I'm fine with merging this PR, but I would not actually build container images for Darwin.

thomastaylor312: Yeah, no need to build a container image here. Let me rename the file real quick

thomastaylor312: Ok, file has been renamed!

pohly:
/assign @jsafrane

For approval.

thomastaylor312:
Thanks all!

PR #171, opened 2021-10-19
feat: add an option to enable pprof profiling

aramase:
Signed-off-by: Anish Ramasekar <anish.ramasekar@gmail.com>

**What type of PR is this?**
/kind feature

- Adds an option to enable pprof profiling

Fixes https://github.com/kubernetes-csi/node-driver-registrar/issues/170

Starting the `node-driver-registrar` container with `--enable-pprof=true` will enable a debug http endpoint at `--pprof-port` (default: 6065). Accessing this will also require `port-forward`:

```bash
kubectl port-forward csi-secrets-store-secrets-store-csi-driver-7x44t node-driver-registrar 6065:6065 &
curl localhost:6065/debug/pprof
```

```release-note
Added an option to enable pprof profiling with --enable-pprof. The profiling server port can be configured using --pprof-port.
```

### DIFF
--- a/cmd/csi-node-driver-registrar/main.go
+++ b/cmd/csi-node-driver-registrar/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	_ "net/http/pprof"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -66,9 +67,10 @@ var (
 	pluginRegistrationPath  = flag.String("plugin-registration-path", "/registration", "Path to Kubernetes plugin registration directory.")
 	kubeletRegistrationPath = flag.String("kubelet-registration-path", "", "Path of the CSI driver socket on the Kubernetes host machine.")
 	healthzPort             = flag.Int("health-port", 0, "(deprecated) TCP port for healthz requests. Set to 0 to disable the healthz server. Only one of `--health-port` and `--http-endpoint` can be set.")
-	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
+	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including pprof and the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
 	showVersion             = flag.Bool("version", false, "Show version.")
 	mode                    = flag.String("mode", ModeRegistration, `The running mode of node-driver-registrar. "registration" runs node-driver-registrar as a long running process. "kubelet-registration-probe" runs as a health check and returns a status code of 0 if the driver was registered successfully, in the probe definition make sure that the value of --kubelet-registration-path is the same as in the container.`)
+	enableProfile           = flag.Bool("enable-pprof", false, "enable pprof profiling")
 
 	// Set during compilation time
 	version = "unknown"
--- a/cmd/csi-node-driver-registrar/node_register.go
+++ b/cmd/csi-node-driver-registrar/node_register.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"os/signal"
 	"runtime"
@@ -68,7 +69,7 @@ func nodeRegister(csiDriverName, httpEndpoint string) {
 	// Registers kubelet plugin watcher api.
 	registerapi.RegisterRegistrationServer(grpcServer, registrar)
 
-	go healthzServer(socketPath, httpEndpoint)
+	go httpServer(socketPath, httpEndpoint)
 	go removeRegSocket(csiDriverName)
 	// Starts service
 	if err := grpcServer.Serve(lis); err != nil {
@@ -86,14 +87,16 @@ func buildSocketPath(csiDriverName string) string {
 	return fmt.Sprintf("%s/%s-reg.sock", *pluginRegistrationPath, csiDriverName)
 }
 
-func healthzServer(socketPath string, httpEndpoint string) {
+func httpServer(socketPath string, httpEndpoint string) {
 	if httpEndpoint == "" {
-		klog.Infof("Skipping healthz server because HTTP endpoint is set to: %q", httpEndpoint)
+		klog.Infof("Skipping HTTP server because endpoint is set to: %q", httpEndpoint)
 		return
 	}
-	klog.Infof("Starting healthz server at HTTP endpoint: %v\n", httpEndpoint)
+	klog.Infof("Starting HTTP server at endpoint: %v\n", httpEndpoint)
 
-	http.HandleFunc("/healthz", func(w http.ResponseWriter, req *http.Request) {
+	// Prepare http endpoint for healthz + profiling (if enabled)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, req *http.Request) {
 		socketExists, err := util.DoesSocketExist(socketPath)
 		if err == nil && socketExists {
 			w.WriteHeader(http.StatusOK)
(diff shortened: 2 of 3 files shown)


verult:
Hi @aramase, could you document how you tested this change in the PR description?

cmd/csi-node-driver-registrar/main.go
```diff
@@ -69,6 +71,8 @@ var (
 	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
 	showVersion             = flag.Bool("version", false, "Show version.")
 	mode                    = flag.String("mode", ModeRegistration, `The running mode of node-driver-registrar. "registration" runs node-driver-registrar as a long running process. "kubelet-registration-probe" runs as a health check and returns a status code of 0 if the driver was registered successfully, in the probe definition make sure that the value of --kubelet-registration-path is the same as in the container.`)
+	enableProfile           = flag.Bool("enable-pprof", false, "enable pprof profiling")
+	profilePort             = flag.Int("pprof-port", 6065, "port for pprof profiling")
```
verult, line 75: I suggest using `http-endpoint` instead, similar to controller sidecars: https://github.com/kubernetes-csi/external-provisioner/blob/7ae61466a611061ba70da2679358c8bfc8546b6f/cmd/csi-provisioner/csi-provisioner.go#L85

Future HTTP handlers can reuse this endpoint. With that format both `enable-pprof` and `http-endpoint` need to be set explicitly.

aramase: @verult Not sure I understand! The `http-endpoint` is used explicitly for healthcheck and can be optionally used for metrics in the `node-driver-registrar` too. `pprof-port` is explicitly for pprof. Is there a need to reuse http handlers? Let me know what you think and I can make the changes. 

https://github.com/kubernetes-csi/node-driver-registrar/blob/c63de8714b1b5df76e3310b6224c814de6e11438/cmd/csi-node-driver-registrar/main.go#L68-L69

aramase:
> Hi @aramase, thanks for the PR! Could you document how you tested this change in the PR description?

Added to the PR description.

cmd/csi-node-driver-registrar/main.go
verult: `http-endpoint` was intended to be the single endpoint serving all HTTP requests in each controller sidecar, because we don't want each handler to require a different port. It's may not be only for health check and metrics - it just so happens that today those are the necessary handlers, but there could be others in the future.

In the case of node-driver-registrar, following the controller sidecar pattern, `http-endpoint` would also be the single http endpoint, but the only handler it serves is pprof.

Another reason to do this, other than consistency, is to avoid having to deprecate the `pprof-port` flag when we need to add a new handler.

Hopefully it's not too complicated to set up a common mux and add the pprof handler. Let me know what you think.

aramase: sounds good! Let me update it to use the `http-handler`.

andyzhangx (approved):
/lgtm

cmd/csi-node-driver-registrar/main.go
aramase: @verult Added a common mux. Let me know what you think.

cmd/csi-node-driver-registrar/main.go
```diff
@@ -69,6 +70,7 @@ var (
 	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
```
verult, line 70: Add pprof in flag description

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -110,6 +113,15 @@ func healthzServer(socketPath string, httpEndpoint string) {
 		}
 	})
 
+	if *enableProfile {
+		klog.InfoS("Starting profiling", "endpoint", httpEndpoint)
+
+		mux.HandleFunc("/debug/pprof/", pprof.Index)
```
verult, line 119: Does Cmdline also need to be here?

verult (changes requested):
LGTM! Some minor comments

cmd/csi-node-driver-registrar/node_register.go
aramase: Added `Cmdline`

cmd/csi-node-driver-registrar/main.go
verult: To make the sentence flow better, maybe something like "...diagnostics, including *pprof and* the health check..."

verult:
Added just one last comment

cmd/csi-node-driver-registrar/main.go
aramase: Done!

aramase:
> Added just one last comment

Thanks, updated the flag description.

verult:
/lgtm

/assign @msau42 
for approval

cmd/csi-node-driver-registrar/main.go
```diff
@@ -66,9 +67,10 @@ var (
...
 	kubeletRegistrationPath = flag.String("kubelet-registration-path", "", "Path of the CSI driver socket on the Kubernetes host machine.")
 	healthzPort             = flag.Int("health-port", 0, "(deprecated) TCP port for healthz requests. Set to 0 to disable the healthz server. Only one of `--health-port` and `--http-endpoint` can be set.")
-	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
+	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including pprof and the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
```
msau42, line 70: @verult it looks like if you set httpEndpoint, it always enables the health check. Would we potentially want to enable the health check and pprof independently? Or maybe it doesn't matter since pprof is for debugging anyway.

cmd/csi-node-driver-registrar/main.go
```diff
@@ -66,9 +67,10 @@ var (
...
 	showVersion             = flag.Bool("version", false, "Show version.")
 	mode                    = flag.String("mode", ModeRegistration, `The running mode of node-driver-registrar. "registration" runs node-driver-registrar as a long running process. "kubelet-registration-probe" runs as a health check and returns a status code of 0 if the driver was registered successfully, in the probe definition make sure that the value of --kubelet-registration-path is the same as in the container.`)
+	enableProfile           = flag.Bool("enable-pprof", false, "enable pprof profiling")
```
msau42, line 73: Just for my curiosity, can you elaborate what issue you are seeing where you want to enable pprof? I would have expected that other sidecars like provisioner or attacher would experience much more load than node-driver-registrar.

cmd/csi-node-driver-registrar/main.go
verult: pprof can still be disabled by a separate flag, so the only case this prevents is having pprof on but health check off. I think it's OK not to allow that case since the health check is fairly lightweight. If we need to allow that case we could add a flag in the future.

cmd/csi-node-driver-registrar/main.go
aramase: In the latest node-driver-registrar release [v2.3.0](https://github.com/kubernetes-csi/node-driver-registrar/releases/tag/v2.3.0) with the addition of liveness probe check, I'm seeing the container consume more memory in case of linux and windows. This is a considerable increase compared to v2.2.0 release. Adding this flag to make it easier for debugging this scenario. Also, we should maybe add this to the other sidecars as well?

msau42: Yes I think this could be useful to enable for other sidecars too. Can we quickly double check that the other sidecars would be able to use the same cli option? I would like to make sure we try to keep args consistent with other sidecars as much as possible.

verult: Yeah the other sidecars that have `http-endpoint` also use a `ServeMux`. We could register the handler the same way and gate it by an `enable-pprof` switch.

e.g. https://github.com/kubernetes-csi/external-provisioner/blob/7ae61466a611061ba70da2679358c8bfc8546b6f/cmd/csi-provisioner/csi-provisioner.go#L496

msau42:
/approve

Thanks!

cmd/csi-node-driver-registrar/main.go
aramase: Let me know if we want to add this option in all the CSI sidecars. I would be happy to open a PR.

msau42: Yes I think that would be valuable, thanks!

PR #214, opened 2022-08-05
Remove mkdir call while creating the registration probe file

### LINKED ISSUE #213: "Failed to create registration probe file" with readOnlyRootFilesystem
When using `readOnlyRootFilesystem: true` the `node-driver-registrar` tries to recursively create the directory the probe file is in, leading to an error:

```
E0804 18:06:27.375117       1 main.go:107] "Failed to create registration probe file" err="mkdir /var/lib/kubelet: read-only file system" registrationProbePath="/var/lib/kubelet/plugins/ebs.csi.aws.com/registration"
```

In this scenario `/var/lib/kubelet/plugins/ebs.csi.aws.com/` is a `hostPath` mounted into the container (and registration works just fine), but `node-driver-registrar` tries to create `/var/lib/kubelet/` despite the fact it already exists and fails, never creating the probe file.

Semi-related: see https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1333#issuecomment-1205658890

mauriciopoppe:
**What type of PR is this?**
>
> /kind bug

node-driver-registrar fails to come up on a filesystem where the `/var/lib/kubelet` directory is read only

Under the assumption that the `kubeletRegistrationPath` path exists we don't need to create subdirectories, we will still create the probe exec file by default though.

Fixes #213

```release-note
The directories in `kubeletRegistrationPath` are assumed to be already created and are no longer created by node-driver-registrar
```

/cc @jingxu97 @msau42

### DIFF
--- a/pkg/util/util_unix.go
+++ b/pkg/util/util_unix.go
@@ -22,7 +22,6 @@ package util
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"golang.org/x/sys/unix"
 )
@@ -88,11 +87,6 @@ func TouchFile(filePath string) error {
 		return err
 	}
 	if !exists {
-		err := os.MkdirAll(filepath.Dir(filePath), 0755)
-		if err != nil {
-			return err
-		}
-
 		file, err := os.Create(filePath)
 		if err != nil {
 			return err
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -28,7 +28,6 @@ import (
 )
 
 var socketFileName = "reg.sock"
-var kubeletRegistrationPath = "/var/lib/kubelet/plugins/csi-dummy/registration"
 
 // TestSocketFileDoesNotExist - Test1: file does not exist. So clean up should be successful.
 func TestSocketFileDoesNotExist(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pkg/util/util_unix.go
```diff
@@ -88,11 +87,6 @@ func TouchFile(filePath string) error {
 		return err
 	}
 	if !exists {
-		err := os.MkdirAll(filepath.Dir(filePath), 0755)
```
msau42, line 91: Is this assumption true for all csi drivers?

mauriciopoppe: The path for the temp registration file is hidden from the users, the steps to create it are:

- take the kubelet registration path: `/var/lib/kubelet/plugins/<drivername.example.com>/csi.sock`
- on a successful registration, write the temp file to the same dirname: `/var/lib/kubelet/plugins/<drivername.example.com>/registration`

I think the directory `/var/lib/kubelet/plugins/<drivername.example.com>/` is created in advance in the driver spec e.g. from the README

```
      volumes:
        - name: plugin-dir
          hostPath:
            path: /var/lib/kubelet/plugins/<drivername.example.com>/
            type: DirectoryOrCreate
```

I checked a few drivers:

- [aws ebs](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/4b64c4e1b1a252c94e5940bcc588c9233c03f3d1/deploy/kubernetes/base/node.yaml#L113)
- [pd csi](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/7a126c25d05ffe04feecf950e0b0fb89e4d7af67/deploy/kubernetes/base/node_linux/node.yaml#L79)
- [azuredisk](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/7b3b87cb3420f0051c128495ea2a01bb7e1ddf38/deploy/v1.21.0/csi-azuredisk-node.yaml#L144)

I think the assumption that this directory is created is true therefore node-driver-registrar should only create the file assuming that the dir already exists

PR #251, opened 2022-11-24
Use golang.org/x/text@v0.4.0 to mitigate CVE-2022-32149

### LINKED ISSUE #249: v2.6.1 is affected by CVE-2022-32149
The latest release is affected by vulnerability CVE-2022-32149 which is marked as high severity. The issue is being addressed in other parts of Kubernetes as well, see: https://github.com/kubernetes/kubernetes/pull/112989/files

My proposed solution is to add the following directive to go.mod:
```
replace golang.org/x/text => golang.org/x/text v0.4.0
```

I'd be happy to submit a PR when the solution is agreed upon.

knopt:
Signed-off-by: Tomasz Knopik <tknopik@nvidia.com>

**What type of PR is this?**
/kind bug

Fixes CVE-2022-32149 which is high severity vulnerability discovered in node-driver-registrar dependencies.

Fixes #249 

```
$  cat go.sum | grep golang.org/x/text
golang.org/x/text v0.4.0 h1:BrVqGRd7+k1DiOgtnFvAkoQEWQvBc25ouMJM6429SFg=
golang.org/x/text v0.4.0/go.mod h1:mrYo+phRRbMaCq/xk9113O4dZlRixOauAjOtrjsXDZ8=
```

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,3 +48,5 @@ require (
 	k8s.io/component-base v0.25.4 // indirect
 	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1 // indirect
 )
+
+replace golang.org/x/text => golang.org/x/text v0.4.0
--- a/go.sum
+++ b/go.sum
@@ -396,6 +396,7 @@ github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9de
 github.com/yuin/goldmark v1.1.32/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
+github.com/yuin/goldmark v1.4.13/go.mod h1:6yULJ656Px+3vBD8DxQVa3kxgyrAnzto9xy5taEt/CY=
 go.etcd.io/bbolt v1.3.2/go.mod h1:IbVyRI1SCnLcuJnV2u8VeU0CEYM7e686BmAb1XKL+uU=
 go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=
 go.opencensus.io v0.22.0/go.mod h1:+kGneAE2xo2IficOXnaByMWTGM9T73dGwxeWcUqIpI8=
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -48,3 +48,5 @@ require (
 	k8s.io/component-base v0.25.4 // indirect
 	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1 // indirect
 )
+
+replace golang.org/x/text => golang.org/x/text v0.4.0
```
msau42, line 52: Why do we need the replace?

msau42: I see, it's a transitive dependency that has that requirement. I guess the potential concern is that version combination being untested.

mauriciopoppe: There's also an analysis that Jan did in https://kubernetes.slack.com/archives/C8EJ01Z46/p1669292513960349

knopt: To address the compatibility issues, we could change replace clause from v0.4.0 to v0.3.8. However, according to semantic versioning, which is followed by golang, upgrade of MINOR is backward compatible.

The analysis seems to be correct, but upgrading seems like a more trustworthy approach.

xing-yang:
@knopt Have you scanned the image after this fix?  Do you see the CVE's gone after this?

go.mod
msau42: We discussed this at today's csi standup. It looks like go.sum lists all the possible versions that could be used, but that doesn't necessarily mean that the version is being used. So cve scanners that rely on go.sum may produce a lot of false positives.

https://stackoverflow.com/questions/70128040/content-of-go-sum-and-modules-really-used-by-a-go-application
https://stackoverflow.com/questions/69825533/why-does-go-sum-include-so-many-older-packages

Given that this change removed the entries from go.sum, but didn't actually change anything in the vendor directory (which reflects what's actually being used) confirms this.

What scanner are you using? It looks like trivy no longer relies on go.sum when the go version in go.mod is >= 1.17: https://github.com/aquasecurity/trivy/issues/1973. But dependabot still does not: https://github.com/dependabot/dependabot-core/issues/4740

knopt: I used a commercial security analysis tool to detect vulns. It is a good point that presence of a dependency in go.sum does not necessarily mean that it's actually used. However, the affected version (v0.3.7) is explicitly listed in `go.mod` of node-driver-registrar (indirect dependency but still a dependency) (it's not go.sum issue that was mentioned).

I agree that it looks like the affected part of the code is not used (even indirectly) by node-driver-registrar. However, afaik security scanners standards are not designed to go into analysis on a function level, they stop at reporting affected package version. Yet, they are still useful tool.

I believe that there is a choice to be made here by maintainers. Every time a new vulnerability is discovered either affected dependency should be updated regardless if the node-driver-registrar is using the affected, or do the vulnerability security audit every time and somehow maintain list of such vulnerabilities which are believed to be not affecting the repo and make all users aware that their scans might produce higher-than-usual number of false-positives.

msau42: In the go.mod, it looks like only 0.4.0 is listed: https://github.com/kubernetes-csi/node-driver-registrar/pull/251/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R42

Anyway, I think we should go ahead and merge this because it is a burden to have users manage exceptions in their scanners, but we should also notify the scanning tools of false-positives so that they can be improved.

knopt: Yeah, you are actually right, I was looking at 2.6.1 and 2.6.2 tags. `0.4.0` is not on master b/c of https://github.com/kubernetes-csi/node-driver-registrar/pull/243 which I've just noticed. 

The above makes this PR redundant, thank you for the discussion though! I think it's important to recognize how vulns should be addressed.

Issue #315, opened 2023-07-21
Unnecessary write to root filesystem


mattcary:
cmd/main.go:GetInfo writes to registrationProbePath, even when --mode=registration. But kubelet-registration-path refers to /var/lib/kubelet, which is not mounted into the registrar container in --mode=registration.

/assign @mattcary

mauriciopoppe:
This is related to https://github.com/kubernetes-csi/node-driver-registrar/issues/309, I wrote some strategies there about how to make it work by making the --mode=kubelet-registration-info always return exit code 0 and removing all the other supporting code including the unnecessary write to the root fs.

mattcary:
I think it is not correct to use --kubelet-registration-path in mode=registration, as it's a container local path and not actually /var/lib/kubelet. So we should be able to make that change immediately? Let me start a PR to be concrete about what I mean.

@jsafrane

mauriciopoppe:
yes, we can do the changes for #309 immediately i.e. https://github.com/kubernetes-csi/node-driver-registrar/issues/309#issuecomment-1606178000

mattcary:
Oh, I see what you mean. Thanks for clarifying. I'll do that.

mauriciopoppe:
#316 implemented https://github.com/kubernetes-csi/node-driver-registrar/issues/315#issuecomment-1646058503. With the new release the write no longer to the root filesystem no longer happens.

PR #316, opened 2023-07-21
Deprecate kubelet-registration-probe

### LINKED ISSUE #309: Deprecate `--mode` option `kubelet-registration-probe`
In https://github.com/kubernetes-csi/node-driver-registrar/issues/143 we added a workaround for a race condition happening in Windows only in the kubelet registration mechanism, as part of the workaround I added the flag mode in https://github.com/kubernetes-csi/node-driver-registrar/pull/152 which through a livenessProbe would crash the Pod if the kubelet registration failed, fortunately @luckerby found the race in https://github.com/kubernetes/kubernetes/issues/104584#issuecomment-1108492823 and fixed it upstream.

With the above fix the `mode` flag is no longer needed and can be deprecated, its introduction created issues with the probe file it created in readOnly filesystems as seen in https://github.com/kubernetes-csi/node-driver-registrar/issues/244.

I still need to define the deprecation timeline and the steps to remove the flag.

```[tasklist]
### Tasks
- [ ] Remove code that creates the probe
- [ ] Update the docs removing references of this flag
```

As an alternative we can keep the feature but under the hood we would do nothing when `--mode=kubelet-registration-probe`, if so then I can safely remove the code that tries to create a file.

mattcary:
This also removes an unnecessary write to the root filesystem which prevents this driver as running under a readOnlyRootFilesystem security context

Fixes #309, #315

/assign @mauriciopoppe 

/kind cleanup

```release-note
Deprecate kubelet-registration-probe and remove unnecessary write to root filesystem.
```

### DIFF
--- a/cmd/csi-node-driver-registrar/main.go
+++ b/cmd/csi-node-driver-registrar/main.go
@@ -22,11 +22,9 @@ import (
 	"fmt"
 	_ "net/http/pprof"
 	"os"
-	"path/filepath"
 	"strconv"
 	"time"
 
-	"github.com/kubernetes-csi/node-driver-registrar/pkg/util"
 	"k8s.io/klog/v2"
 
 	"github.com/kubernetes-csi/csi-lib-utils/connection"
@@ -44,18 +42,9 @@ const (
 )
 
 const (
-	// ModeRegistration runs node-driver-registrar as a long running process
-	ModeRegistration = "registration"
-
 	// ModeKubeletRegistrationProbe makes node-driver-registrar act as an exec probe
 	// that checks if the kubelet plugin registration succeeded.
 	ModeKubeletRegistrationProbe = "kubelet-registration-probe"
-)
-
-var (
-	// The registration probe path, set when the program runs and used as the path of the file
-	// to create when the kubelet plugin registration succeeds.
-	registrationProbePath = ""
 )
 
 // Command line flags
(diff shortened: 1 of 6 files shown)


cmd/csi-node-driver-registrar/main.go
```diff
@@ -69,7 +60,7 @@ var (
...
 	httpEndpoint            = flag.String("http-endpoint", "", "The TCP network address where the HTTP server for diagnostics, including pprof and the health check indicating whether the registration socket exists, will listen (example: `:8080`). The default is empty string, which means the server is disabled. Only one of `--health-port` and `--http-endpoint` can be set.")
 	showVersion             = flag.Bool("version", false, "Show version.")
-	mode                    = flag.String("mode", ModeRegistration, `The running mode of node-driver-registrar. "registration" runs node-driver-registrar as a long running process. "kubelet-registration-probe" runs as a health check and returns a status code of 0 if the driver was registered successfully, in the probe definition make sure that the value of --kubelet-registration-path is the same as in the container.`)
+	mode                    = flag.String("mode", ModeRegistration, "DEPRECATED. This flag is no longer used. If this is set to kubelet-registration-probe, the driver will immediately exist successfully without registering with CSI.")
```
mauriciopoppe, line 63:
```suggestion
	mode                    = flag.String("mode", ModeRegistration, "DEPRECATED. This flag is no longer used. If this is set to kubelet-registration-probe, the driver will immediately exit successfully without registering with CSI.")
```

cmd/csi-node-driver-registrar/main.go
```diff
@@ -101,14 +92,6 @@ func newRegistrationServer(driverName string, endpoint string, versions []string
 func (e registrationServer) GetInfo(ctx context.Context, req *registerapi.InfoRequest) (*registerapi.PluginInfo, error) {
 	klog.Infof("Received GetInfo call: %+v", req)
 
-	// on successful registration, create the registration probe file
-	err := util.TouchFile(registrationProbePath)
```
mauriciopoppe, line 105: There's test code for this function that we can remove

cmd/csi-node-driver-registrar/main.go
```diff
@@ -145,22 +128,10 @@ func main() {
 		klog.Error("kubelet-registration-path is a required parameter")
 		os.Exit(1)
 	}
-	// set after we made sure that *kubeletRegistrationPath exists
-	kubeletRegistrationPathDir := filepath.Dir(*kubeletRegistrationPath)
-	registrationProbePath = filepath.Join(kubeletRegistrationPathDir, "registration")
 
-	// with the mode kubelet-registration-probe
 	if modeIsKubeletRegistrationProbe() {
-		lockfileExists, err := util.DoesFileExist(registrationProbePath)
```
mauriciopoppe, line 154: I forgot if I added this function or if it was already there, there's test code for it that might be dead code too

mauriciopoppe (commented):
Thanks for the help removing the code, could you also update README.md please?

cmd/csi-node-driver-registrar/main.go
mattcary: Well, it will immediately exist too... :-P

cmd/csi-node-driver-registrar/main.go
mattcary: Oops I didn't even think about tests, thx.

cmd/csi-node-driver-registrar/main.go
mattcary: +1

mattcary:
Thanks for the review, I'll update the README which I also overlooked.

cmd/csi-node-driver-registrar/main.go
mattcary: Yup it seems to be dead.

mauriciopoppe:
hey @mattcary, friendly ping about the node-driver-registrar change about the kubelet-registration-probe

mattcary:
Sorry for the delay. Done.

PR #322, opened 2023-08-11
Check if registration socket is still valid

gnufied:
Health check should check if registration socket is really responsive.

```release-note
Do not merely check presence of registration socket, but also check if it responds to grpc requests
```


cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +136,43 @@ func httpServer(socketPath string, httpEndpoint string) {
...
+
+	klog.V(1).Infof("Calling node registrar to check if it still responds")
+	ctx, cancel := context.WithTimeout(context.Background(), *operationTimeout)
+	defer cancel()
+
+	client := registerapi.NewRegistrationClient(csiConn)
+
+	infoRequest := &registerapi.InfoRequest{}
+
+	info, err := client.GetInfo(ctx, infoRequest)
```
jsafrane, line 157: `Probe()` would be more appropriate

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +136,43 @@ func httpServer(socketPath string, httpEndpoint string) {
 	klog.Fatal(http.ListenAndServe(httpEndpoint, mux))
 }
 
+func checkLiveRegistrationSocket(socketFile, csiDriverName string) error {
+	klog.V(1).Infof("Attempting to open a gRPC connection with: %q", socketFile)
+	cmm := metrics.NewCSIMetricsManagerForSidecar("")
```
jsafrane, line 141: The manager should not be re-created for each health check. Please use the one from main.go

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +136,43 @@ func httpServer(socketPath string, httpEndpoint string) {
...
+	csiConn, err := connection.Connect(socketFile, cmm)
```
jsafrane, line 142: it's not CSI connection, it's registration socket

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +136,43 @@ func httpServer(socketPath string, httpEndpoint string) {
...
+	if err != nil {
+		return fmt.Errorf("error connecting to node-registrar socket %s: %v", socketFile, err)
+	}
+
+	defer closeGrpcConnection(socketFile, csiConn)
```
jsafrane, line 147: Does it make sense to close the connection? All CSI sidecars re-use the same connection for all requests. gRPC does some recovery when the socket is closed, AFAIK.

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +136,43 @@ func httpServer(socketPath string, httpEndpoint string) {
...
+
+	klog.V(1).Infof("Calling node registrar to check if it still responds")
```
jsafrane, line 149: V(2)

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +136,43 @@ func httpServer(socketPath string, httpEndpoint string) {
...
+
+	info, err := client.GetInfo(ctx, infoRequest)
+	if err != nil {
+		return fmt.Errorf("error getting info from node-registrar socket: %v", err)
+	}
+
+	if info.Name == csiDriverName {
+		return nil
+	}
+	return fmt.Errorf("invalid driver name %s", info.Name)
```
jsafrane, line 165: Will this be useful? Did we ever saw CSI drivers changing their name?

cmd/csi-node-driver-registrar/node_register.go
gnufied: for sidecars I guess it makes sense. I am less sure about periodic checks. I think closing the connection here is fine.

cmd/csi-node-driver-registrar/node_register.go
gnufied: I do not think driver name should change without requiring re-registration right?

cmd/csi-node-driver-registrar/node_register.go
gnufied: I removed need for metrics manager. I think it was unncessary for this use case anyways.

cmd/csi-node-driver-registrar/node_register.go
gnufied: renamed the socket file.

cmd/csi-node-driver-registrar/node_register.go
```diff
@@ -126,6 +135,42 @@ func httpServer(socketPath string, httpEndpoint string) {
...
+func checkLiveRegistrationSocket(socketFile, csiDriverName string) error {
+	klog.V(2).Infof("Attempting to open a gRPC connection with: %q", socketFile)
+	grpcConn, err := connection.ConnectWithoutMetrics(socketFile)
+	if err != nil {
+		return fmt.Errorf("error connecting to node-registrar socket %s: %v", socketFile, err)
+	}
+
+	defer closeGrpcConnection(socketFile, grpcConn)
+
+	klog.V(1).Infof("Calling node registrar to check if it still responds")
```
jsafrane, line 147: nit: V(2)

mowangdk:
We have a similar problem on our environment, csi-node-driver-registrar process still exists, but the socket has been deactivated, what is the root cause of the problem?  Is it an OS issue?

PR #323, opened 2023-08-17
Update dependency go modules for k8s v1.28.0

Sneha-at:
Ran kubernetes-csi/csi-release-tools go-get-kubernetes.sh -p 1.28.0


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/kubernetes-csi/csi-lib-utils v0.15.0-rc0.0.20230811202321-0f6938005f63
 	golang.org/x/sys v0.11.0
 	google.golang.org/grpc v1.57.0
-	k8s.io/client-go v0.28.0-rc.0
+	k8s.io/client-go v0.28.0
 	k8s.io/klog/v2 v2.100.1
 	k8s.io/kubelet v0.28.0-rc.0
```
jsafrane, line 11: This should be v0.28.0

Sneha-at: Updated the version

go.mod
```diff
@@ -6,9 +6,9 @@ require (
 	github.com/kubernetes-csi/csi-lib-utils v0.15.0-rc0.0.20230811202321-0f6938005f63
```
sunnylovestiramisu, line 6: This is still rc,  we need to update when csi-lib-utils v0.15.0 is ready

Sneha-at: Updated `csi-lib-utils v0.15.0`

PR #555, opened 2026-02-20
Update release-tools

sunnylovestiramisu:
**What type of PR is this?**
/kind failing-test

The old logcheck in release-tools with 0.8 version is using pkg/mod/golang.org/x/tools@v0.21.0 which is compatible with golang 1.24+, lockcheck fails if we build on top of 1.24+. While the master branch of release tools have been updated to log check 0.10+, we need the sub branch to be updated.

Fixes #

```release-note
NONE
```


sunnylovestiramisu:
Since it does not work, we just disable the test internally

/close

